PR #32, opened 2017-01-13
Preserve metadata in gzip_dir

shyamd:
Copy the metadata over to the gzip'ed file using shutil copystat

### DIFF
--- a/monty/shutil.py
+++ b/monty/shutil.py
@@ -53,13 +53,14 @@ def gzip_dir(path, compresslevel=6):
             GzipFile, 6 is default for gzip.
     """
     for f in os.listdir(path):
+        full_f = os.path.join(path, f)
         if not f.lower().endswith("gz"):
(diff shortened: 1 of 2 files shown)


shyamd:
Primary use case is custodian's gzip option to preserve the modification timestamp.

shyuep:
Pls add unittest.

shyamd:
Added in a Unit test and fixed some path resolution issues in gzip_dir

shyuep:
TEsts failed. Think you need to do AlmostEqual.

shyamd:
So it seems on Mac filestamps are resolved to the second level so assertEqual was working fine on my mac, but technically they can be resolved to less than a second, so AlmostEqual is the correct way. The nuances of modern computing...

shyuep:
Note that you still have an issue with the text you read from the file. This seems that you were using python 2.7. I recommend that you move on to Py3. I will fix it this time.

PR #34, opened 2017-10-25
Change out of the temp directory before deleting

jesperkristensen58:
On Windows you get an error if calling shutil.rmtree(...) before os.chdir(...). If switching the order it works.
Linux can handle this.

## Summary

<Short few sentences, and summary of the major changes in bullet 
points>

* Feature 1
* Feature 2
* Fix 1
* Fix 2

## TODO

<If this is a work-in-progress, write something about what else needs to be 
done>

* Feature 1 supports a, but not b.


shyuep:
Thanks!

PR #35, opened 2018-02-13
Allow binary in jsanitize for bson

shyamd:
Simple fix to allow binary data if allow_bson == True in jsanitize

### DIFF
--- a/monty/json.py
+++ b/monty/json.py
@@ -258,7 +258,7 @@ def jsanitize(obj, strict=False, allow_bson=False):
     Returns:
         Sanitized dict that can be json serialized.
     """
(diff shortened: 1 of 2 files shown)


shyuep:
Pls fix the failing travis test.

shyamd:
The test is failing in python2 because future doesn't seem to be installed . Do we want to scuttle py2 compatibility as a whole or add that into the requirements?

shyuep:
Does the same functionality work in six? I have six?

shyamd:
Had to remove the test for how python stringiness binary data. It's too fundamentally different in python2 and python3 without explicitly checking for version.

PR #43, opened 2019-03-26
Use basename in determination of output/input type in monty serialization

JosephMontoya-TRI:
I encountered a somewhat weird error recently when using monty serialization in that temporary filenames generated by both standard library tempfile and monty's tempfile tools would occasionally have "mpk" in the absolute paths of filenames.  This would result in inconsistent (albeit somewhat rare) errors in integration tests when I was using temporary directories to test file output routines.  

Example:
```
dumpfn({"var_1": 1}, "tmp123MpK/output.json")
```

I dunno if it's really worth it, but this might be avoided by inferring the serialization format from a basename of the filename, rather than just the filename, which could include the path.

### DIFF
--- a/monty/serialization.py
+++ b/monty/serialization.py
@@ -58,7 +58,7 @@ def loadfn(fn, *args, **kwargs):
     Returns:
         (object) Result of json/yaml/msgpack.load.
     """
-    if "mpk" in fn.lower():
+    if "mpk" in os.path.basename(fn).lower():
         if msgpack is None:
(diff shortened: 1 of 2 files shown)


shyuep:
unittest?

shyuep:
Also, current code is failing travis tests.

JosephMontoya-TRI:
Added discriminating test and fixed the unimported reference.

shyuep:
Thanks.

PR #47, opened 2019-08-16
Updates to dumpfn/loadfn

clegaspi:
## Summary

This PR adds support for "yml" as an extension for YAML files in `loadfn` and `dumpfn`. The PR also changes the logic for determining file type in these functions by requiring that the extensions be preceded by a period character to match the appropriate file type (i.e. ".mpk", ".yaml", ".yml"), because "yml" may come up in the file name outside of the extension.

Tests caught a bug when opening json files stored as gzip with python<3.6. The default read format used by `zopen` is binary, and `json.loads()` does not support binary strings until python 3.6. So, the PR now explicitly sets read mode to text (`'rt'`).

Tests have been expanded to verify these functions work with the file extensions/types and documentation has been updated.

Per a discussion with colleagues, I thought it might be useful to have a keyword argument to manually select the file format like: `dumpfn(obj, 'somefile.txt', fmt='yaml')`. Would have to check the kwargs for json/yaml/msgpack.load to make sure there are no clashes. Thoughts? (I have included this as the last commit, but can revert if objectionable.)

* Add support for YAML files with ".yml" extension in `dumpfn`/`loadfn`
* Tweak file format determination logic in `dumpfn`/`loadfn` to require a period before the extension
* Add manual format specification through `fmt` kwarg to `dumpfn`/`loadfn`

## TODO

* Update changelog (is this done manually in the rst file?)

### DIFF
--- a/monty/serialization.py
+++ b/monty/serialization.py
@@ -47,9 +47,11 @@ __date__ = '7/29/14'
 def loadfn(fn, *args, **kwargs):
     """
     Loads json/yaml/msgpack directly from a filename instead of a
-    File-like object. For YAML, ruamel.yaml must be installed. The file type is
-    automatically detected. YAML is assumed if the filename contains "yaml"
-    (lower or upper case). Otherwise, json is always assumed.
+    File-like object. File may also be a BZ2 (".BZ2") or GZIP (".GZ", ".Z") compressed file.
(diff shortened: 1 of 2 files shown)


shyuep:
I have no objections to adding fmt. ANd thanks for the contributions. I will merge.

PR #48, opened 2019-08-16
Redirect Monty loading

shyamd:
This PR provides a redirect for `monty` to be able to load classes from a new location after coding changes

# TODO:
Test the loading from .mson.yaml

@shyue, any suggestions on how to design a test for loading from a central user file? Also how do you feel about the settings loading code as it is?


shyuep:
Ok, not quite sure what this is meant to do, but for tests involving central user files, you can write the test to create those files temporarily and then do the test and clean up afterwards.

shyamd:
Ah! This is supposed to supply a redirect for class definition lookup for MontyDecoder similar to what is in Fireworks right now. If you move the python code for a class, you can just put the old location/new location as a key-value pair in your `~/.monty.yaml` and it will update the dictionary on the fly to be able to still load the object.

shyuep:
I see. Ok. Let me know when it is ready to merge.

shyamd:
Should be all set.

PR #67, opened 2019-11-01
[WIP] Check for valid kwargs in default from_dict

mkhorton:
Improves robustness of monty when backwards incompatible changes are made. See https://github.com/materialsproject/custodian/pull/130

### DIFF
--- a/monty/json.py
+++ b/monty/json.py
@@ -15,6 +15,8 @@ from importlib import import_module
 
 from inspect import getfullargspec
 
+from warnings import warn
+
 try:
     import numpy as np
 except ImportError:
@@ -163,8 +165,22 @@ class MSONable:
         :param d: Dict representation.
         :return: MSONable class.
         """
+        
+        # only pass valid args to class
+        args = getfullargspec(cls).args
         decoded = {k: MontyDecoder().process_decoded(v) for k, v in d.items()
-                   if not k.startswith("@")}
+                   if (k in args) and (not k.startswith("@"))}
+        
+        # if d contains args that are not valid, warn user (may be due to deprecation)
+        invalid_keys = {k for k in d.keys() if (k not in args) and (not k.startswith("@"))}
+        if invalid_keys:
+            msg = "Trying to de-serialize dictionary into {}, "
+                  "contains invalid key(s): {}.".format(cls.__name__, invalid_keys)
+            if "@version" in d:
+                msg += " Dictionary was serialized using {} version {}."\
+                    .format(d["@module"], d["@version"])
+            warn(msg)
+                
         return cls(**decoded)
 
     def to_json(self) -> str:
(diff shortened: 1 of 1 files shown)


shyuep:
I am unclear why this is a warning message is needed. Surely if the deserialization fail it is quite clear that the argument is wrong.

mkhorton:
The case where this came up:

1. Class `A` has argument `X` and is serialized into a dictionary `d`.
2. Code is updated, and argument `X` is removed. User tries to de-serialize `d` using default `from_dict`. De-serialization fails because argument `X` no longer exists.
3. De-serialization could still succeed automatically if `X` was just ignored; however, the user should be warned because perhaps this is not what they want.

This came up because one of the Custodian error handlers dropped an argument, but we have thousands of FireWorks that have still that argument hard-coded, so now even some `lpad` commands won't work because it's trying and failing to de-serialize these old handler objects. But it can happen (and has happened) elsewhere too when making backwards-incompatible changes.

shyuep:
It is a very bad idea to deserialize by ignoring arguments. Code should be updated to handle deprecations. If devs want to ignore arguments, they can by all means write a wrapper function to clean the dict before doing deserialization. But it should not be that Monty just ignores bad arguments.

mkhorton:
So there is a pull request in Custodian to handle the deprecation in this specific instance -- but in general, the design of FireWorks is such that it becomes very fragile when these kinds of backwards incompatible changes are made. I think a warning is a fair compromise (imo).

For example, currently any historical data that has been serialized via `dumpfn` becomes completely unreadable if one of these arg changes happens without surgical modification of the file.

I agree that devs should handle this properly in the first place before making such changes but this does not always happen ...

shyuep (changes requested):
**Automated review by Claude (on behalf of @shyuep)**

The intent here — warning when a serialized dict contains keys unknown to the current class constructor — is good for catching version-mismatch bugs. However, there is a correctness issue with the current implementation.

**Behavioral regression for `**kwargs` constructors:**
`getfullargspec(cls).args` only captures explicitly named parameters. If a class's `__init__` accepts `**kwargs`, all extra keys would be silently dropped rather than forwarded. Previously, `from_dict` passed all non-`@` keys through, which is the correct behavior for such classes. The filter should be skipped (or handle `varkw`) when `getfullargspec(cls.__init__).varkw is not None`.

**Merge conflicts:**
The PR has merge conflicts against master and needs to be rebased. Given that it has been open since 2019, a rebase would be needed regardless.

Suggested fix sketch:
```python
spec = getfullargspec(cls.__init__)
if spec.varkw is None:
    decoded = {k: ... for k, v in d.items() if k in spec.args and not k.startswith("@")}
    # warn on invalid_keys
else:
    decoded = {k: ... for k, v in d.items() if not k.startswith("@")}
```

Please address these and rebase onto master.

---

PR #73, opened 2019-11-20
Fix GZip not crawling sub-directories

shyamd:
## Summary

`gzip_dirs` was having problems with sub-directories. It was trying to gzip them which causes an exception and fails. This updated method crawls the sub-directories and gzip's files in them.

I refrained from using `Pathlib` to preserve compatibility, but I think at some point `monty` should switch over as it is SO nice.

### DIFF
--- a/monty/shutil.py
+++ b/monty/shutil.py
@@ -3,7 +3,6 @@ Copying and zipping utilities. Works on directories mostly.
 """
 
 import os
-import glob
 import shutil
 import warnings
 from gzip import GzipFile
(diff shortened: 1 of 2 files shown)


shyuep:
Instead of using glob, which is fragile, how about you use os.walk? That will crawl the entire tree.

I agree that Pathlib is nice. We can actually go ahead and implement using Pathlib (though in this case, perhaps os.walk is still a better solution). From 2020, we are going full Py3 in monty. All major codes are dropping Py2 support anyway.

Issue #89, opened 2019-12-20
Support for tuples, namedtuples, OrderedDict and possibly others


davidwaroquiers:
## Summary
In some objects, tuples, namedtuples, OrderedDict and other data-container objects are used. Currently, MontyEncoder and MontyDecoder do not support them (it actually converts tuples to lists, OrderedDicts to dicts)
* When decoding back these objects, they are not anymore the same type of data object and this could be a problem in some cases.
* For OrderedDict, it might not be such of a problem as dicts are now guaranteed to be ordered in python 3.7, if I'm not wrong.

Is there any plan to implement such a feature ? Would it be useful for people ?

## Example code for namedtuple

```python
from collections import namedtuple
from monty.json import MontyEncoder
from monty.json import MontyDecoder

me = MontyEncoder()
md = MontyDecoder()

Point = namedtuple('Point', field_names=['x', 'y'])
a = Point(1, 2)

mydict = {'point': a}
mydecodeddict = md.decode(me.encode(mydict))

print(mydict == mydecodeddict) # Returns False
```

## Suggested solution (if known)

* Deal with these specific data objects in a similar way numpy is handled.

shyuep:
I have no problems with this. Feel free to implement and submit a PR.

PR #100, opened 2020-01-14
WIP Added support for tuples, namedtuples (both from collections and typing), sets, frozensets and OrderedDict's in MSONable, MontyEncoder and MontyDecoder

davidwaroquiers:
## Summary
Solves  #89 

* tuple's, collections.namedtuple's, typing.NamedTuple's, set's, frozenset's and collections.OrderedDict's are automatically serialized as special dictionaries, that allow to reconstruct the proper object upon deserialization.

## TODO

Might want to add support for other objects, e.g. Data classes and possibly others.
Not sure whether the last try-except in the default method of MontyEncoder is still necessary as the as_dict is now directly performed (recursively) at the beginning of the encode method of MontyEncoder.


tests/test_collections.py
```diff
@@ -49,5 +52,27 @@ def test_tree(self):
         self.assertEqual(x['a']['b']['c']['d'], 1)
 
 
+def test_is_namedtuple():
+    a_nt = namedtuple('a', ['x', 'y', 'z'])
+    a1 = a_nt(1, 2, 3)
+    assert a1 == (1, 2, 3)
+    assert is_namedtuple(a1) is True
```
shyuep: I don't think you need "is True". The assert tests whether it is True.
Similarly, 3 lines later you just need "assert not is_named_tuple..."

tests/test_json.py
```diff
@@ -188,6 +191,44 @@ def test_enum_serialization(self):
         self.assertEqual(e_new.name, e.name)
         self.assertEqual(e_new.value, e.value)
 
+    def test_tuple_serialization(self):
+        a = GoodMSONClass(a=1, b=tuple(), c=1)
+        afromdict = GoodMSONClass.from_dict(a.as_dict())
+        assert type(a.b) is tuple
```
shyuep: I think isinstance is a better approach. This applies to all instances where you use type(X) is or == something.

tests/test_collections.py
davidwaroquiers: It is actually not exactly equivalent. The "is True" also checks that the returned value is a bool and not for example an empty list vs a non empty list. If is_namedtuple(a1) were to return, e.g. ["hello"] (or to return [] in the "False" case), "assert is_namedtuple(a1)" would also pass while the API would have changed. I would keep the "is True" if that's ok for you. Same for 3 lines after.

tests/test_json.py
davidwaroquiers: isinstance checks that the instance is a tuple or a subclass of tuple. I think it is more safe to use type(a.b) is tuple as it would prevent someone (a weirdo btw ;-) to change the MSONable as_dict and/or from_dict to return a subclass of tuple when deserializing. What do you think ?

tests/test_collections.py
shyuep: Yes, I know it is not exactly equivalent, but we need to think of typical use cases. When people use that method, they are not going to check that it is a boolean type. They will merely check that it is a "True-like" value. E.g., if is_namedtuple(a1). That's the Pythonic way of doing this.

tests/test_json.py
shyuep: I agree. But this is a test, not the actual code. So here, it suffices to make sure it is an instance of tuple.

davidwaroquiers: If the user serializes a tuple, he should expect to get back exactly a tuple when deserializing. Using isinstance in the test allows to change that. I could modify the MSONable from_dict to return a subclass of a tuple instead of a real tuple.

shyuep:
Is this still being actively worked on or should I close this?

davidwaroquiers:
> Is this still being actively worked on or should I close this?

Hi @shyuep 


Thanks for the heads up on this. Completely forgot about it. I guess that the question was more rhetorical and that "actively worked on" was ironic since it's been 5 years :)


I'll see quickly if we still have a use case for this (as I forgot about it and did not pursue, probably not, at least not urgently). I'll close it myself end of next week if there is no need to keep it.


Best

shyuep (changes requested):
**Automated review by Claude (on behalf of @shyuep)**

This PR adds useful serialization support for tuples, namedtuples, sets, and OrderedDicts, but has several blocking issues that need to be resolved before it can be merged.

**Critical — Python 3.9 incompatibility:**
The entire `is_NamedTuple` detection and `validate_NamedTuple` rely on `_field_types`, which was deprecated in Python 3.8 and **removed in Python 3.9**. As a result, `is_NamedTuple` will always return `False` on Python 3.9+, and `validate_NamedTuple` will raise `AttributeError`. The correct modern equivalent is `__annotations__`. The distinguishing logic between `collections.namedtuple` and `typing.NamedTuple` needs to be rewritten for Python 3.9+.

**Title vs. implementation mismatch:**
The PR title and description mention `frozenset` support, but `frozenset` is not handled anywhere in the diff.

**Stale version guards:**
Monty already requires Python ≥ 3.8, so all the `sys.version_info >= (3, 6)` / `>= (3, 7)` checks are dead code and should be removed.

**Test style:**
Using `exec()` to define classes in tests (to work around Python version restrictions) is now unnecessary given the minimum Python version. These should be replaced with regular class definitions.

**Merge conflicts:**
The PR has merge conflicts against the current master and will need to be rebased.

The overall approach is sound — please update to handle modern Python and rebase onto master.

---

PR #118, opened 2020-03-18
travis: add all officially supported python versions

Mic92:
## Summary

We saw tests break on python3.7. According to setup.py python3.7/python3.8 are officially supported and should be therefore also tested.

## TODO

-

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: python
 python:
   - "3.5"
   - "3.6"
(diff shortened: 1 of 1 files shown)


shyuep:
Travis passed... so there are no breaking tests. Thanks for adding the new python versions.
Note that some tests require certain things to be installed/. E.g., msgpack. But these do not have any effect on functionality unless you use certain specific modules.

PR #121, opened 2020-04-06
Update test_serializaiton for latest msgpack

jonringer:
msgpack>=1 returns a str, not bytes.

previous error:
```
    File "/build/source/tests/test_serialization.py", line 57, in <dictcomp>
      self.assertEqual(d, {k.decode('utf-8'): v.decode('utf-8')
  AttributeError: 'str' object has no attribute 'decode'
```


shyuep:
Thanks.

PR #152, opened 2020-07-09
Feature: Add Varidac Args to MSONable

shyamd:
## Summary

This PR adds varidac argument support to Monty MSONable.


shyuep:
Thanks.

PR #166, opened 2020-09-02
Update setup.py

triplepoint:
## Summary
Indicate that monty is only compatible with Python versions 3.5 and later.

That python 3-only code was added to this project without this indicator was allowing pip to install python-3-only code in python 2 environments.  This is still a problem, as versions still exist in PYPI that are Python-3-only and which don't have this attribute, but better late than never.


shyuep:
Thanks.

PR #239, opened 2021-04-14
Add support for serialising functions, methods, & classes

utf:
## Summary

Added support for serialising callables. This covers builtins, functions, classes, unbound methods (staticmethods), and bound methods (instance and class methods). Bound instance methods are only serialisable if the instance itself can be serialised (i.e., it is MSONable).

I've made sure that callable MSONable instances (i.e., MSONable classes that define the `__call__` method) are still serialised as an instance rather than as a callable.

### Format

The serialisation format is:

```python
d = {
    "@module": <module>,
    "@callable": <function or class name>,
    "@bound": <serialized bound object>
}
```

## Examples

The entire `data` array can now be serialised using `MontyEncoder` or using `jsanitize`.

```python
import os
from enum import Enum


class MyClass(MSONable):

    def __init__(self, a):
        self.a = a

    def method(self):
        pass

    @staticmethod
    def staticmethod(self):
        pass

    @classmethod
    def classmethod(cls):
        pass

    class NestedClass:
        def inner_method(self):
            pass


def my_callable(a, b):
    return a + b


instance = MyClass(1)
data = [
    # builtins
    str,
    list,

    # functions
    os.path.join,
    my_callable,

    # unbound methods
    MyClass.NestedClass.inner_method,
    MyClass.staticmethod,
    instance.staticmethod,

    # methods bound to classes
    MyClass.classmethod,
    instance.classmethod,

    # classes
    MyClass,
    Enum,  # using Enum as an example, but any class (MSONable or not) will work

    # instance methods
    instance.method
]
```


shyuep:
I am unclear why this is needed?

utf:
For atomate2, we have developed a new framework for writing workflows. We have moved away from writing Firetasks to using direct serialisation of function calls. This has been implemented in the jobflow library (https://github.com/hackingmaterials/jobflow) currently under development.

As part of the library, there is the option to impose a schema on calculation outputs. This is implemented by passing a pydantic model (basically just a class) which is then applied to the outputs.

Both of these features (delayed initialisation of a class and delayed function calls) require serialising functions and classes. I think that these are useful additions for monty and significantly expand the scope of objects that can be serialised.

PR #251, opened 2021-05-10
Allow serialisation of builtin functions

utf:
## Summary

My previous PR (#239) added support for serialising functions, methods, & classes. Builtin primitives (`str`, `int`, etc) and  classes (`list`, `set`, etc) also worked but builtin functions (`open`, `sum`, etc) did not.

This PR extends support for builtin functions.

### DIFF
--- a/monty/json.py
+++ b/monty/json.py
@@ -5,6 +5,7 @@ JSON serialization and deserialization utilities.
 
 import os
 import json
+import types
 import datetime
 
 from hashlib import sha1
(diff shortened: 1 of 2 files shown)


utf:
There are several tools to serialise builtin functions. The most popular example is pickle. For example, the following works:

```python
import pickle
pickle.dumps(sum)
```

The downside is that pickle is platform dependent. I actually see this PR as making the serialisation capabilities of monty more in line with more established serialisation tools. The benefit is that the implementation in monty can be used to serialise/deserialise on any system.

As for codes that use this functionality in practice - well there are lots of examples. Maybe the most used is `dask`. For example, see the documentation on futures: https://docs.dask.org/en/latest/futures.html In their examples you can replace their `add` function with `sum` and it will still work. A similar approach is taken with Ray: https://ray.io

In fact, the new framework for writing workflows in atomate2 is heavily inspired by the futures module of dask. The general concept is that a function is serialised and then executed later. This PR actually completes the feature spec of monty, meaning that it can now serialise all functions, rather than just methods and non-builtin functions.

PR #265, opened 2021-06-09
Add support for serializing pydantic models

utf:
## Summary

Pydantic models are basically just fancy dictionaries. There is already partial support for pydantic models in monty, I just added support for serialising and deserialising them.

Note, I had to edit the pylintrc to work with pydantic. This is because pylint doesn't work well with c extensions as described here: https://github.com/samuelcolvin/pydantic/issues/1961


shyuep:
Thanks.

PR #267, opened 2021-06-10
Add support for numpy generic types

utf:
## Summary

Added support for numpy generic types (np.float32, np.int64, etc) in jsanitize. They were already supported in `MontyEncoder`.


utf:
Thanks very much. Would it be possible to release a new version when you get the chance?

shyuep:
Done.

utf:
Fantastic, thank you very much.

PR #272, opened 2021-06-16
Enable serialisation of Enum values

utf:
## Summary

I added an option to jsanitize that allows enums to be serialized as their values (default is False to match the current behaviour). There is already custom logic in the emmet version of jsanitize that does this behaviour but I'd like to see this functionality in monty itself.

However, this is adding another option to jsanitize, so I will understand if you'd rather not merge this.


shyuep:
thanks.

PR #280, opened 2021-07-08
Add pandas DataFrame serialization

mkhorton:
## Summary

This follows the pattern of `numpy`, `bson`, etc. support in `monty`, adding support for `pandas`.

Given that pandas is now in wide-spread use, as numpy, it seems sensible to support this in monty for the purposes of loadfn, dumpfn etc., as well as to avoid custom as_dict, from_dict for classes that have DataFrame attributes.

This uses the native pandas `to_json()` serialization but uses `MontyEncoder` so that any custom data types (like pymatgen Structure) within the DataFrame are also serialized and de-serialized correctly.

This is to address a need in pymatgen (https://github.com/materialsproject/pymatgen/pull/2191). Tagging @htz1992213 @rkingsbury.


htz1992213:
@shyuep  This is a really important feature! I am wondering when this PR can be reviewed and merged. Thanks!!

PR #308, opened 2021-10-13
pprint_json

jmmshn:
## Added interactive tree printing for dictionaries and objects.

Function to recursively display the data in a dictionary.
If an object is used as the argument it will try to deserialize the object first. before display.
The weird way it does exceptions is just to make sure it shows everything it can show and keeps going instead of erroring out.
Since most of the time just seeing a portion of the data is enough.

Example usage:
image


jmmshn:
Still not sure why pylint is failing for this

PR #325, opened 2021-12-01
Adds support for lzma/xz format in zopen

zhubonan:
Allow xz/lzma format to be used with `zopen`.

## Summary

Adds support for xz/lzma format with `zopen`. 

The `xz` format offers higher compression ratio than `gzip` but takes much longer time to compression.
However, decompression is nearly as fast. Tests on common VASP output text files shows up to 40% more
size reduction compared to `gzip`.


shyuep:
Thanks.

PR #335, opened 2022-01-01
Fix decoding of dictionaries with @module and @callable but no @class

arosen93:
## Summary
As described in more detail in #336, calling `monty.json.MontyDecoder.process_decoded` on a dictionary with `"@module"` and `"@callable"` (but not `"@class"`) keys results in the following error: `UnboundLocalError: local variable 'classname' referenced before assignment`.

This PR resolves the issue by setting `classname=None` and simply returning `{self.process_decoded(k): self.process_decoded(v) for k, v in d.items()}` in this scenario.

The PR closes issue #336, and I've confirmed that the tests still pass. Don't let the block of red in the diff deceive you -- it's almost all from an indent.

### DIFF
--- a/monty/json.py
+++ b/monty/json.py
@@ -364,6 +364,7 @@ class MontyDecoder(json.JSONDecoder):
             elif "@module" in d and "@callable" in d:
                 modname = d["@module"]
                 objname = d["@callable"]
+                classname = None
                 if d.get("@bound", None) is not None:
                     # if the function is bound to an instance or class, first
                     # deserialize the bound object and then remove the object name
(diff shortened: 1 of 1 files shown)


arosen93:
Hi @shyuep, would you mind merging this PR when you get a chance and releasing a new version of monty? This fix is needed for virtually all of my [Jobflow](https://github.com/materialsproject/jobflow) workflows, and it'd be nice to switch to the main branch rather than my fork. Thanks!

shyuep:
Thanks. Merged and new version released.

arosen93:
Thanks!!

PR #358, opened 2022-03-08
Integration of `orjson` for (much) faster JSON encoding and decoding

munrojm:
## Summary

This PR switches monty from using the standard JSON `encode` and `decode` functions, to using the equivalent in the `orjson` package (see https://github.com/ijl/orjson). This greatly speeds up both `MontyDecoder` and `MontyEncoder`.  

One important note is that I elected to keep the fact that MontyDecoder/Encoder subclasses `json.JSONDecoder` and `json.JSONEncoder` respectively. I think this effectively communicates that both custom classes can still be used with json.dumps/loads as the passed `cls` argument. 

I will also point out that this change was motivated by slowness issues in the new Materials Project API from decoding large objects.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,38 +5,9 @@ default_language_version:
 
 ci:
   autoupdate_schedule: monthly
-  skip: [flake8, autoflake, mypy]
+  skip: [flake8, mypy]
 
 repos:
-  - repo: https://github.com/psf/black
-    rev: 22.1.0
-    hooks:
-      - id: black
-
-  - repo: https://github.com/PyCQA/isort
-    rev: 5.10.1
-    hooks:
-      - id: isort
-        args: ["--profile", "black"]
-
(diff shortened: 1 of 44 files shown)


shyuep:
I don't have a problem with this, but it should be made an optional alternative rather than a requirement. In other words, the code should check if orjson is installed and use that if it is, and if not, fall back to standard json.

munrojm:
> I don't have a problem with this, but it should be made an optional alternative rather than a requirement. In other words, the code should check if orjson is installed and use that if it is, and if not, fall back to standard json.

Sounds good, I will make that change.

munrojm:
Re-writing a new `encode` function with recursive pre-processing before passing to `orjson` turned out to be slower than the existing solution. It is better to use the native `orjson` API with the `MontyEncoder.default` method for encoding, which I have added to the docstring. I have kept the optional `orjson` decoding in, as it is trivial to incorporate without any issues.

Additionally, I have fixed some of the CI issues with pre-commit and pylint. This should be good to go now.

PR #363, opened 2022-03-20
ScratchDir improvements: fix symlink removal + new kwargs: `gzip_on_exit` and `delete_if_removed`

arosen93:
Closes #363.

This PR seeks to improve the functionality of `monty.tempfile.ScratchDir`. While I know python has its own `tempfile.TemporaryDirectory`, there are many nice features of `ScratchDir` (e.g. symbolic link making) that warrant its continued use, and I've continually found myself in need of some additional flexibility in its behavior. I outline these changes below.

1. I added a line that removes the symlink on exit. Currently, it doesn't seem to be removed, which isn't ideal because then there is a dead (broken) symlink in the main directory. Anytime you go to run a second job in the same directory, a new symlink then can't be made.
2. I added a `gzip_on_exit` kwarg that will `gzip` the files in the scratch dir before copying them back if `True`. Defaults to `False` for backwards compatibility. While in principle the user could just gzip the files themselves afterwards, in practice this isn't ideal because we typically want to gzip the files while they're in the tmpdir since that may be a location with better I/O performance.
3. I added a `delete_if_removed` kwarg that if `False`, will _not_ delete files in the cwd if they were removed in the tmpdir. The default is `True` for backwards compatibility. I chose to include this for security reasons. If something unexpected were to happen to the files in the tmpdir (whether due to a coding error or some other unexpected issue), then it's possible the source files would be removed too. This can lead to some dangerous consequences, and I wanted to provide the user the option to avoid the concern altogether. A separate motivation was that I was also having some issues with the file deletion on Windows in some edge-cases, but I couldn't pinpoint the problem unfortunately. This resolved it.

There was also never a test for the symlink creation. I have now added that, along with tests for the new features.


arosen93:
@shyuep -- this is good to go if you're okay with it. I've battle-tested it now for the last two weeks as well.

PR #379, opened 2022-04-18
Force `MontyDecoder` to fall back to python JSON decoder when `orjson` errors

munrojm:
## Summary

This PR forces `MontyDecoder` to try and fall back to the native python JSON decoder when an `orjson.JSONDecodeError` is thrown. This addresses instances where special float types like `float.NaN` exist in the target data. These are handled fine by the native python library, but not `orjson` as it takes a stricter approach to conforming to JSON schema standards.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,7 +41,7 @@ repos:
         args: ["--profile", "black"]
 
   - repo: https://github.com/psf/black
-    rev: 22.1.0
+    rev: 22.3.0
     hooks:
       - id: black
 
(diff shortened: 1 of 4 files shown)


mkhorton:
Important context here is that NaN value types are out of spec with the JSON standard, which is why `orjson` doesn't support them (see `allow_nan` in the Python standard library [`json.dump`](https://docs.python.org/3/library/json.html#json.dump) which defaults to `True`, so by default Python will output out-of-spec JSON if NaN values are present). This also applies to (+/-)Infinity.

Having this PR merged into `monty` would be helpful to cover this edge case, so we can continue to benefit from `orjson`'s massive speed improvements while not losing support for files with NaN.

mkhorton:
@shyuep would it be possible for this to be merged/released if approved?

PR #426, opened 2022-08-28
fix a small typo

kmu:
## Summary

Fix a small typo to correctly print the error message.

Before: `and kwargs to be present undera self.kwargs variable`

After:  `and kwargs to be present under a self.kwargs variable`

## TODO

N.A.


shyuep:
Thanks.

PR #444, opened 2022-10-15
Add generic types to @cached_class

janosh:
Solves the issue reported in https://github.com/microsoft/pylance-release/issues/3478 that `@cached_class` obscures the methods on a decorated class from auto-completion and doc string lookup (at least in VS Code and presumably most other IDEs).


shyuep:
I had to revert thise. It was causing test failures.

janosh:
Ah, tests were passing locally and it looks like it's a Python version mismatch. Tests should be fine after importing future annotations.

PR #491, opened 2023-04-09
Fix handling of `datetime` in `jsanitize`

arosen93:
Currently, the `jsanitize()` command in `monty` throws an error if `strict=True` and a `datetime` object is supplied. It's better if we just handle `datetime` objects directly in `jsanitize` given how often they crop up anyway.


arosen93:
@shyuep --- this is ready to merge at your convenience. Thanks!

shyuep:
Thanks Andrew!

PR #536, opened 2023-08-07
Return path to decompressed file from `decompress_file()`

janosh:
Needed for https://github.com/materialsproject/pymatgen/pull/3192#issuecomment-1668420484.

### DIFF
--- a/monty/shutil.py
+++ b/monty/shutil.py
@@ -1,6 +1,5 @@
-"""
-Copying and zipping utilities. Works on directories mostly.
-"""
+"""Copying and zipping utilities. Works on directories mostly."""
+from __future__ import annotations
 
 import os
 import shutil
@@ -20,7 +19,6 @@ def copy_r(src, dst):
         src (str): Source folder to copy.
         dst (str): Destination folder.
     """
-
     abssrc = os.path.abspath(src)
     absdst = os.path.abspath(dst)
     try:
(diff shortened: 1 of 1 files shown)


arosen93:
@janosh: This seems like it was perhaps a breaking change. Could you take a look at https://github.com/Quantum-Accelerators/quacc/actions/runs/5790005449/job/15692210083?pr=660 and see if it seems related?

arosen93:
@janosh: Yeah, actually, seems like this introduce a bug. If the file extension isn't in BZ2, GZ, or Z then `out_file` is never referenced and an error is raised.

monty/shutil.py
```diff
@@ -90,12 +88,12 @@ def compress_dir(path, compression="gz"):
...
             "bz2". Defaults to gz.
     """
-    for parent, subdirs, files in os.walk(path):
+    for parent, _subdirs, files in os.walk(path):
         for f in files:
             compress_file(os.path.join(parent, f), compression=compression)
 
 
-def decompress_file(filepath):
+def decompress_file(filepath) -> str:
```
arosen93, line 96: @janosh: Change this to `str | None`

monty/shutil.py
```diff
@@ -104,14 +102,20 @@ def decompress_file(filepath):
...
     toks = filepath.split(".")
     file_ext = toks[-1].upper()
     if file_ext in ["BZ2", "GZ", "Z"]:
-        with open(".".join(toks[0:-1]), "wb") as f_out, zopen(filepath, "rb") as f_in:
+        out_file = ".".join(toks[0:-1])
+        with zopen(filepath, "rb") as f_in, open(out_file, "wb") as f_out:
             f_out.writelines(f_in)
         os.remove(filepath)
 
+    return out_file
```
arosen93, line 117: @janosh: move this `return` inside the `if` statement so that it only returns the string if it's decompressing a file. Otherwise, have it `return None` (as before).

PR #537, opened 2023-08-07
Fix `decompress_file()` if no compressed file found

janosh:
Thanks @arosen93 https://github.com/materialsvirtuallab/monty/pull/536#issuecomment-1668644134.


arosen93:
Thanks for the quick fix!

PR #548, opened 2023-08-30
Pydantic v2 support for MSONable types

munrojm:
## Summary

This PR adds support for pydantic >= 2.0 with the addition of custom class validation methods to `MSONable`:
- `__get_pydantic_json_schema__` to define pydantic field JSON schema
- `__get_pydantic_core_schema__` to define pydantic field core schema

The same validation function is used for both versions.

### DIFF
--- a/monty/json.py
+++ b/monty/json.py
@@ -227,7 +227,7 @@ class MSONable:
             )
 
     @classmethod
-    def validate_monty(cls, __input_value, _):
+    def validate_monty(cls, __input_value, *agrs, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


arosen93:
Thanks, @munrojm! Very much looking forward to getting this merged. I'm not sure how picky @shyuep is with the codecov tests, but you're super close to the diff hit so if you need to you can probably add another test to get the lovely green check :)

shyuep:
Merged. The codecov is just meant to check if something really has no tests. Here we are well covered.

shyuep:
Already done.

PR #551, opened 2023-09-04
Allow pathlib.Path types in monty.shutil

### LINKED ISSUE #550: Various functions are not safe to use with `pathlib.Path` objects (only `str`)
Several of the functions in monty are not safe to use with `pathlib.Path` objects as the input. This is not a bug since the type-hint or docstring indicates it's just `str`, but it's worth updating at some point.

For instance, take the example below:

https://github.com/materialsvirtuallab/monty/blob/ae7f481a8519b87499e9a09b01ba1df4026ee361/monty/shutil.py#L96-L107

Calling `.split()` on a `Path` object isn't possible.

For maximum flexibility, a simple enhancement would be to call `str(<input>)` at the top of the various path handling functions, which would then allow the input type to be either `str` or `Path`.

arosen93:
Closes #550 by allowing `pathlib.Path` object types in `monty.shutil` functions.


arosen93:
@shyuep: Ready for your review.

shyuep:
Thanks!

PR #582, opened 2023-10-20
Add support for serializing ase Atoms objects

utf:
This PR adds support for serializing ase `Atoms` objects.

We've found this might be a useful feature to have in atomate2. But I think it could be useful more broadly. E.g., `Atoms` objects are used in the MD interface in `matgl`, in `chgnet`, and `quacc` (developed by @Andrew-S-Rosen). Although I understand if this sort of feature is outside the scope of monty.

### DIFF
--- a/monty/json.py
+++ b/monty/json.py
@@ -59,6 +59,11 @@ try:
 except ImportError:
     torch = None  # type: ignore
 
+try:
+    import ase
+except ImportError:
+    ase = None  # type: ignore
+
 __version__ = "3.0.0"
 
 
@@ -374,6 +379,9 @@ class MontyEncoder(json.JSONEncoder):
                     "data": o.to_json(default_handler=MontyEncoder().encode),
                 }
 
+        if ase is not None and isinstance(o, ase.Atoms):
+            return {"@module": "ase", "@class": "Atoms", "data": MontyEncoder().encode(o.todict())}
+
         if bson is not None:
             if isinstance(o, bson.objectid.ObjectId):
                 return {"@module": "bson.objectid", "@class": "ObjectId", "oid": str(o)}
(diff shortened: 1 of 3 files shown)


shyuep:
I agree this is useful. But shouldn't this be implemented in ASE itself? Eg pymatgen implements Structure serialization. Monty is not exactly a materials science code. It is strange to have ASE serialization here.

Andrew-S-Rosen:
ASE does have its own `.todict()` and `.fromdict()` methods. That said, it doesn't currently work if there are constraints. So, I usually have something like the following in my code.

```python
from ase.io.jsonio import decode, encode

def atoms_as_dict(s: Atoms) -> dict:
    # Uses Monty's MSONable spec
    # Normally, we would want to this to be a wrapper around atoms.todict() with @module and
    # @class key-value pairs inserted. However, atoms.todict()/atoms.fromdict() does not currently
    # work properly with constraints.
    return {"@module": "ase.atoms", "@class": "Atoms", "atoms_json": encode(s)}


def atoms_from_dict(d: dict) -> Atoms:
    # Uses Monty's MSONable spec
    # Normally, we would want to have this be a wrapper around atoms.fromdict()
    # that just ignores the @module/@class key-value pairs. However, atoms.todict()/atoms.fromdict()
    # does not currently work properly with constraints.
    return decode(d["atoms_json"])
```

Not sure of the pros/cons compared to your approach @utf. I'd need to investigate. But I just wanted to share what I already know about the topic.

shyuep:
I would be fine adding support for todict and fromdict as alternatives to as_dict and from_dict. Though it would be better if ASE simply adds these aliases.

shyuep:
Closing this for now.

Andrew-S-Rosen:
I know this is quite old, but I wanted to chime in to say that there is basically 0% chance such methods will be able to be added to ASE. It either has to happen in `monty` or monkeypatched in each code that wants to make `Atoms` an `MSONable`.

> I would be fine adding support for todict and fromdict as alternatives to as_dict and from_dict.

Did we want to revisit this proposition? I will open a PR with tests, and you can accept or reject.

I will note that the method proposed by @utf is not ideal. The below will crash because `FixAtoms` is not JSON serializable. 

```python
from ase.build import bulk
from ase.constraints import FixAtoms
from monty.json import MontyEncoder

atoms = bulk("Cu")
c = FixAtoms([0])
atoms.set_constraint(c)
{"@module": "ase", "@class": "Atoms", "data": MontyEncoder().encode(atoms.todict())}
```

[My approach](https://github.com/materialsvirtuallab/monty/pull/582#issuecomment-1772959295) using ASE's JSON encoder/decoder will always work, but it has the downside that the `Atoms` object data is stored as a JSON string, which is not readily parseable via a database query. You'd have to rehydrate it with an `Atoms.from_dict()` call before use.

PR #608, opened 2024-01-25
lazy load pandas

mturiansky:
Addresses https://github.com/materialsproject/pymatgen/issues/3563?notification_referrer_id=NT_kwDOAIdMALI5MTI0NjgwNTQ0Ojg4NjY4MTY#issue-2087635103

This could be implemented for the other optional dependencies, but pandas is the only one that I have installed in my venv that affects the performance of pymatgen.


mturiansky:
I think it should be fixed now. It says "this branch has no conflicts with the base branch" on my end.

shyuep:
Great. Thanks,

Issue #610, opened 2024-01-28
[Bug]: 602 Enum serialization was a breaking change


janosh:
### Version

2024.1.26

### Which OS(es) are you using?

- [x] MacOS
- [x] Windows
- [x] Linux

### What happened?

`TestPhaseDiagram.test_read_json` started failing after https://github.com/materialsproject/pymatgen/commit/5137111a9ed7252013d68fc4c7cea23b5a39ff8a bumping `monty` to `2024.1.26`.

### Code snippet

```python
def test_read_json(self):
    dumpfn(self.pd, f"{self.tmp_path}/pd.json")
    pd = loadfn(f"{self.tmp_path}/pd.json")
    assert isinstance(pd, PhaseDiagram)
    assert {*pd.as_dict()} == {*self.pd.as_dict()}
```

### Log output

```shell
monty/json.py:517: in process_decoded
    return cls_.from_dict(data)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

d = {'value': 'O'}

    @staticmethod
    def from_dict(d) -> Element:
        """Makes Element obey the general json interface used in pymatgen for
        easier serialization.
        """
>       return Element(d["element"])
E       KeyError: 'element'

/Users/janosh/dev/pmg/pymatgen/core/periodic_table.py:755: KeyError
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Andrew-S-Rosen:
CC @jmmshn

jmmshn:
The same `pymatgen` test now passes on my end.

Andrew-S-Rosen:
@janosh This can now be closed.

janosh:
thanks for the quick fix @jmmshn! 👍

PR #611, opened 2024-01-29
[BUG] Enum subclasses with custom `as_dict`, `from_dict` breaks

jmmshn:
## Small bug in Enum Serialization
Addresses: https://github.com/materialsvirtuallab/monty/issues/610

If an Enum subclass like `Element` has a custom `as_dict`/`from_dict`, then it should be the preferred method for serialization.

The original order of the if statements check for `Enum` before it checked if custom constructors which breaks some tests in pymatgen.

### DIFF
--- a/monty/json.py
+++ b/monty/json.py
@@ -412,8 +412,6 @@ class MontyEncoder(json.JSONEncoder):
         try:
             if pydantic is not None and isinstance(o, pydantic.BaseModel):
                 d = o.dict()
-            elif isinstance(o, Enum):
-                d = {"value": o.value}
             elif (
                 dataclasses is not None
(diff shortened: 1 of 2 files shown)


Andrew-S-Rosen:
Thanks, @jmmshn! FYI @shyuep this should unbreak the current pymatgen failing tests on `master`.

Andrew-S-Rosen:
@shyuep --- can we get a new version of `monty` minted to remove the [version pin in pymatgen](https://github.com/materialsproject/pymatgen/blob/dca98bec837e0186fb3aae4d3e6880a39013945d/setup.py#L32)? 🙏

PR #618, opened 2024-02-08
Add support for keyword-only args

FabiPi3:
## Summary

Add support for keyword-only arguments. Consider the following example class:

```
class KwOnlyTest(MSONable):
    def __init__(self, a, *, b):
        self.a = a
        self.b = b

obj = KwOnlyTest(2, b=3)
```

Looking at the args: `getfullargspec(obj.__class__.__init__)` one gets `FullArgSpec(args=['self', 'a'], varargs=None, varkw=None, defaults=None, kwonlyargs=['b'], kwonlydefaults=None, annotations={})`. Here you can see the difference between `args` and `kwonlyargs`. The old implementation uses only the `args`, resulting in missing `b` in the `as_dict()` representation.

I changed it to use `args` and `kwonlyargs`. This fixes the issue and `b` is now included in the dict representation.

## Checklist

- [ x ] Google format doc strings added. Check with `ruff`.
- [ x ] Type annotations included. Check with `mypy`.
- [ x ] Tests added for new features/fixes.

### DIFF
--- a/monty/json.py
+++ b/monty/json.py
@@ -173,7 +173,6 @@ class MSONable:
             d["@version"] = None  # type: ignore
 
         spec = getfullargspec(self.__class__.__init__)
-        args = spec.args
 
         def recursive_as_dict(obj):
             if isinstance(obj, (list, tuple)):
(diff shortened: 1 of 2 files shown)


shyuep:
Thanks!

PR #619, opened 2024-02-09
Add ASE `Atoms` support for de/encoding

Andrew-S-Rosen:
## Summary

Note: will fix test shortly. 

This is a follow-up to #582 by @utf with the goal of adding support for `Atoms` encoding/decoding via the `MSONable` spec. For pros/cons of this, refer to the original discussion of #582.

@utf: If you have a better mechanism to serialize the `Atoms` that results in fully reversible (de)serialization without relying on storing the data as JSON strings, let me know. The mechanism you had originally proposed in #582 does not work when `Atoms` has non-JSON serializable components, like when a `FixAtoms` constraint is applied.

## Checklist

- [ ] Google format doc strings added. Check with `ruff`.
- [ ] Type annotations included. Check with `mypy`.
- [X] Tests added for new features/fixes.
- [ ] If applicable, new classes/functions/modules have [`duecredit`](https://github.com/duecredit/duecredit) `@due.dcite` decorators to reference relevant papers by DOI ([example](https://github.com/materialsproject/pymatgen/blob/91dbe6ee9ed01d781a9388bf147648e20c6d58e0/pymatgen/core/lattice.py#L1168-L1172))

Tip: Install `pre-commit` hooks to auto-check types and linting before every commit:

```sh
pip install -U pre-commit
pre-commit install
```


shyuep:
I don’t mean to be difficult, but I really think this is not suitable for Monty. It is far too materials science specific while Monty has nothing related to matsci at all. In the end, the guiding principle cannot be that “ASE developers are unwilling to do it, so let’s shove it to some other codebase where it is definitely unsuitable”. At most, I will tolerate some version of it in pymatgen. E.g., we create a subclass of Atoms called PymatgensBetterAtoms that implements as_dict and from_dict and supports MSONAble. This can be used in place of all places that use Atoms. It also has the advantage that we can implement whatever additional functionality we want without having to deal with the whims of other devs.

Andrew-S-Rosen:
> I don’t mean to be difficult, but I really think this is not suitable for Monty. It is far too materials science specific while Monty has nothing related to matsci at all. In the end, the guiding principle cannot be that “ASE developers are unwilling to do it, so let’s shove it to some other codebase where it is definitely unsuitable”. At most, I will tolerate some version of it in pymatgen. E.g., we create a subclass of Atoms called PymatgensBetterAtoms that implements as_dict and from_dict and supports MSONAble. This can be used in place of all places that use Atoms. It also has the advantage that we can implement whatever additional functionality we want without having to deal with the whims of other devs.

In short: I, in fact, agree. I will respond more over email to your other inquiries.

PR #628, opened 2024-02-25
Check for symlinks when using `copy_r` and/or `compress_file`

Andrew-S-Rosen:
If a symlink is present when using `copy_r`, you'll get into an infinite, recursive loop of copying. This PR ignores symlinks when using `copy_r`.

I also fixed a related issue with `compress_file`.


Andrew-S-Rosen:
@shyuep --- this is ready to go! Would be grateful for a release sometime shortly after this is merged, if possible. 🙏

shyuep:
Thanks!

Issue #598, opened 2023-11-24
[Bug]: `DeprecationWarning` due to use of `general_plain_validator_function` in `MSONable.__get_pydantic_core_schema__` with Pydantic v2


ugognw:
https://github.com/materialsvirtuallab/monty/blob/5d37679e1f0cd58a8d165d76f3a095cb71266831/monty/json.py#L291-L300

Pydantic suggests a straight swap for `core_schema.with_info_plain_validator_function`

janosh:
@shyuep Did you mean to assign @munrojm?

DanielYang59:
Thanks! Fixed in #631.

DanielYang59:
@shyuep I believe we can close this issue now. Thanks!

DanielYang59:
@shyuep Can we close this now? Pretty sure this has been fixed. Thanks

PR #631, opened 2024-02-27
Add an optional arg `deadline` to `dev.deprecated` to raise warning after deadline 

DanielYang59:
### Major changes:

- Add an optional arg `deadline` for declaration of the deadline, and raise a CI warning pass that date 
(thanks @janosh for the helpful discussion in https://github.com/materialsproject/pymatgen/pull/3656#discussion_r1502199179)
- Add type hints

### Minor tweaks:

- Remove unused deps in `requirements-ci.txt` (let me know if they're intended otherwise)
- Minor clean up of `test_dev` unit test file (removed a random `Class A` unused anywhere)
- Replace deprecated `general_plain_validator_function` from `monty/json.py`

### Side notes:

- running `pytest` on `test_tempfile.py` seems to modify `test_files/3000_lines.txt.gz` in place, this should not happen.

### DIFF
--- a/docs/monty.dev.md
+++ b/docs/monty.dev.md
@@ -26,19 +26,6 @@ with a possible replacement.
 * **Returns**
   Original function, but with a warning to use the updated class.
 
-## monty.dev.get_ncpus()
-
-**NOTE**: If you are using Python >= 2.7, multiprocessing.cpu_count() already
-provides the number of CPUs. In fact, this is the first method tried.
-The purpose of this function is to cater to old Python versions that
-still exist on many Linux style clusters.
-
-Number of virtual or physical CPUs on this system, i.e.
-user/real as output by time(1) when called with an optimally scaling
-userspace-only program. Return -1 if ncpus cannot be detected. Taken from:
-[http://stackoverflow.com/questions/1006289/how-to-find-out-the-number-of](http://stackoverflow.com/questions/1006289/how-to-find-out-the-number-of)-
-cpus-in-python
-
 ## monty.dev.install_excepthook(hook_type=’color’, \*\*kwargs)
 
 This function replaces the original python traceback with an improved
--- a/docs/monty.dev.md
+++ b/docs/monty.dev.md
@@ -18,7 +18,7 @@ with a possible replacement.
   * **replacement** (*callable*) – A replacement class or method.
   * **message** (*str*) – A warning message to be displayed.
   * **category** (*Warning*) – Choose the category of the warning to issue. Defaults
-    to FutureWarning. Another choice can be DeprecationWarning. NOte that
+    to FutureWarning. Another choice can be DeprecationWarning. Note that
     FutureWarning is meant for end users and is always shown unless silenced.
     DeprecationWarning is meant for developers and is never shown unless
     python is run in developmental mode or the filter is changed. Make
--- a/monty/dev.py
+++ b/monty/dev.py
@@ -72,7 +72,7 @@ def deprecated(
         return wrapped
 
     # Raise a CI error after removal deadline
-    if deadline is not None and "CI" in os.environ and datetime.now() > deadline:
+    if deadline is not None and os.getenv("CI") and datetime.now() > deadline:
         raise DeprecationWarning(
             "This function should have been removed on {deadline:%Y-%m-%d}."
         )
--- a/monty/dev.py
+++ b/monty/dev.py
@@ -27,7 +27,7 @@ def deprecated(
         replacement (callable): A replacement class or method.
         message (str): A warning message to be displayed.
         deadline (Optional[tuple[int, int, int]]): Optional deadline for removal
-            of the old function/class, in format (yyyy, MM, dd). A CI error would
+            of the old function/class, in format (yyyy, MM, dd). A CI warning would
             be raised after this date.
         category (Warning): Choose the category of the warning to issue. Defaults
             to FutureWarning. Another choice can be DeprecationWarning. Note that
@@ -76,7 +76,7 @@ def deprecated(
 
         return wrapped
 
-    # Raise a CI error after removal deadline
+    # Raise a CI warning after removal deadline
     if deadline is None:
         _deadline = None
 
--- a/monty/dev.py
+++ b/monty/dev.py
@@ -5,8 +5,10 @@ particularly useful for developers. E.g., deprecating methods / classes, etc.
 
 import functools
 import logging
+import os
 import sys
 import warnings
+from datetime import datetime
 from typing import Callable, Optional, Type
 
 logger = logging.getLogger(__name__)
@@ -14,7 +16,8 @@ logger = logging.getLogger(__name__)
 
 def deprecated(
     replacement: Optional[Callable] = None,
-    message: Optional[str] = None,
+    message: str = "",
+    deadline: Optional[datetime] = None,
     category: Type[Warning] = FutureWarning,
 ):
     """
@@ -23,6 +26,8 @@ def deprecated(
     Args:
         replacement (callable): A replacement class or method.
         message (str): A warning message to be displayed.
+        deadline (datetime): Optional deadline for removal of the old function/class.
+            A CI error would be raised after this date.
         category (Warning): Choose the category of the warning to issue. Defaults
             to FutureWarning. Another choice can be DeprecationWarning. Note that
             FutureWarning is meant for end users and is always shown unless silenced.
@@ -34,8 +39,17 @@ def deprecated(
         Original function, but with a warning to use the updated class.
     """
 
-    def craft_message(old: Callable, replacement: Callable, message: str):
+    def craft_message(
+        old: Callable,
+        replacement: Callable,
+        message: str,
+        deadline: datetime,
+    ):
         msg = f"{old.__name__} is deprecated"
+
+        if deadline is not None:
+            msg += f", and would be removed on {deadline.strftime('%Y-%m-%d')}\n"
+
         if replacement is not None:
             if isinstance(replacement, property):
                 r = replacement.fget
@@ -44,17 +58,22 @@ def deprecated(
             else:
                 r = replacement
             msg += f"; use {r.__name__} in {r.__module__} instead."
-        if message is not None:
+
+        if message:
             msg += "\n" + message
         return msg
 
     def deprecated_decorator(old: Callable):
         def wrapped(*args, **kwargs):
-            msg = craft_message(old, replacement, message)
+            msg = craft_message(old, replacement, message, deadline)
             warnings.warn(msg, category=category, stacklevel=2)
             return old(*args, **kwargs)
 
         return wrapped
+
+    # Raise a CI error after removal deadline
+    if deadline is not None and "CI" in os.environ and datetime.now() > deadline:
+        raise RuntimeError("This function should have been removed.")
 
     return deprecated_decorator
 
--- a/monty/dev.py
+++ b/monty/dev.py
@@ -48,7 +48,7 @@ def deprecated(
         msg = f"{old.__name__} is deprecated"
 
         if deadline is not None:
-            msg += f", and would be removed on {deadline.strftime('%Y-%m-%d')}\n"
+            msg += f", and will be removed on {deadline:%Y-%m-%d}\n"
 
         if replacement is not None:
             if isinstance(replacement, property):
@@ -73,7 +73,9 @@ def deprecated(
 
     # Raise a CI error after removal deadline
     if deadline is not None and "CI" in os.environ and datetime.now() > deadline:
-        raise RuntimeError("This function should have been removed.")
+        raise DeprecationWarning(
+            "This function should have been removed on {deadline:%Y-%m-%d}."
+        )
 
     return deprecated_decorator
 
--- a/monty/dev.py
+++ b/monty/dev.py
@@ -17,7 +17,7 @@ logger = logging.getLogger(__name__)
 def deprecated(
     replacement: Optional[Callable] = None,
     message: str = "",
-    deadline: Optional[datetime] = None,
+    deadline: Optional[tuple[int, int, int]] = None,
     category: Type[Warning] = FutureWarning,
 ):
     """
@@ -26,8 +26,9 @@ def deprecated(
     Args:
         replacement (callable): A replacement class or method.
         message (str): A warning message to be displayed.
-        deadline (datetime): Optional deadline for removal of the old function/class.
-            A CI error would be raised after this date.
+        deadline (Optional[tuple[int, int, int]]): Optional deadline for removal
+            of the old function/class, in format (yyyy, MM, dd). A CI error would
+            be raised after this date.
         category (Warning): Choose the category of the warning to issue. Defaults
             to FutureWarning. Another choice can be DeprecationWarning. Note that
             FutureWarning is meant for end users and is always shown unless silenced.
@@ -38,6 +39,12 @@ def deprecated(
     Returns:
         Original function, but with a warning to use the updated class.
     """
+
+    def _convert_date(date: tuple[int, int, int]) -> datetime:
+        """Convert a date in int tuple for datetime type.
+        Expect the date in (yyyy, MM, dd) format.
+        """
+        return datetime(*date)
 
     def craft_message(
         old: Callable,
@@ -65,17 +72,22 @@ def deprecated(
 
     def deprecated_decorator(old: Callable):
         def wrapped(*args, **kwargs):
-            msg = craft_message(old, replacement, message, deadline)
+            msg = craft_message(old, replacement, message, _deadline)
             warnings.warn(msg, category=category, stacklevel=2)
             return old(*args, **kwargs)
 
         return wrapped
 
     # Raise a CI error after removal deadline
-    if deadline is not None and os.getenv("CI") and datetime.now() > deadline:
-        raise DeprecationWarning(
-            "This function should have been removed on {deadline:%Y-%m-%d}."
-        )
+    if deadline is None:
+        _deadline = None
+
+    else:
+        _deadline = _convert_date(deadline)
+        if os.getenv("CI") and datetime.now() > _deadline:
+            raise DeprecationWarning(
+                "This function should have been removed on {deadline:%Y-%m-%d}."
+            )
 
     return deprecated_decorator
 
--- a/monty/dev.py
+++ b/monty/dev.py
@@ -41,9 +41,7 @@ def deprecated(
     """
 
     def _convert_date(date: tuple[int, int, int]) -> datetime:
-        """Convert a date in int tuple for datetime type.
-        Expect the date in (yyyy, MM, dd) format.
-        """
+        """Convert date as int tuple (yyyy, MM, dd) to datetime type."""
         return datetime(*date)
 
     def craft_message(
--- a/monty/dev.py
+++ b/monty/dev.py
@@ -7,11 +7,16 @@ import functools
 import logging
 import sys
 import warnings
+from typing import Callable, Optional, Type
 
 logger = logging.getLogger(__name__)
 
 
-def deprecated(replacement=None, message=None, category=FutureWarning):
+def deprecated(
+    replacement: Optional[Callable] = None,
+    message: Optional[str] = None,
+    category: Type[Warning] = FutureWarning,
+):
     """
     Decorator to mark classes or functions as deprecated, with a possible replacement.
 
@@ -19,7 +24,7 @@ def deprecated(replacement=None, message=None, category=FutureWarning):
         replacement (callable): A replacement class or method.
         message (str): A warning message to be displayed.
         category (Warning): Choose the category of the warning to issue. Defaults
-            to FutureWarning. Another choice can be DeprecationWarning. NOte that
+            to FutureWarning. Another choice can be DeprecationWarning. Note that
             FutureWarning is meant for end users and is always shown unless silenced.
             DeprecationWarning is meant for developers and is never shown unless
             python is run in developmental mode or the filter is changed. Make
@@ -29,7 +34,7 @@ def deprecated(replacement=None, message=None, category=FutureWarning):
         Original function, but with a warning to use the updated class.
     """
 
-    def craft_message(old, replacement, message):
+    def craft_message(old: Callable, replacement: Callable, message: str):
         msg = f"{old.__name__} is deprecated"
         if replacement is not None:
             if isinstance(replacement, property):
@@ -43,7 +48,7 @@ def deprecated(replacement=None, message=None, category=FutureWarning):
             msg += "\n" + message
         return msg
 
-    def deprecated_decorator(old):
+    def deprecated_decorator(old: Callable):
         def wrapped(*args, **kwargs):
             msg = craft_message(old, replacement, message)
             warnings.warn(msg, category=category, stacklevel=2)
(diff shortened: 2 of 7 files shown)


requirements-ci.txt
```diff
@@ -1,12 +1,9 @@
 pytest
 pytest-cov
 coverage
-coveralls
 pycodestyle
 mypy
-pydocstyle
 pydantic
-flake8
```
DanielYang59, line 9: Let me know if they're still needed anywhere. Thanks!

monty/dev.py
```diff
@@ -29,8 +39,17 @@ def deprecated(replacement=None, message=None, category=FutureWarning):
...
+    def craft_message(
+        old: Callable,
+        replacement: Callable,
+        message: str,
+        deadline: datetime,
+    ):
         msg = f"{old.__name__} is deprecated"
+
+        if deadline is not None:
+            msg += f", and would be removed on {deadline.strftime('%Y-%m-%d')}\n"
```
janosh, line 51: simplify to

```suggestion
            msg += f", and would be removed on {deadline:%Y-%m-%d}\n"
```

janosh: and (would->will) be removed on

monty/dev.py
```diff
@@ -39,18 +58,23 @@ def craft_message(old, replacement, message):
...
-            msg = craft_message(old, replacement, message)
+            msg = craft_message(old, replacement, message, deadline)
             warnings.warn(msg, category=category, stacklevel=2)
             return old(*args, **kwargs)
 
         return wrapped
 
+    # Raise a CI error after removal deadline
+    if deadline is not None and "CI" in os.environ and datetime.now() > deadline:
+        raise RuntimeError("This function should have been removed.")
```
janosh, line 76:
```suggestion
        raise RuntimeError("This function should have been removed on {deadline:%Y-%m-%d}.")
```

monty/dev.py
```diff
@@ -39,18 +58,23 @@ def craft_message(old, replacement, message):
...
         def wrapped(*args, **kwargs):
-            msg = craft_message(old, replacement, message)
+            msg = craft_message(old, replacement, message, deadline)
             warnings.warn(msg, category=category, stacklevel=2)
             return old(*args, **kwargs)
 
         return wrapped
 
+    # Raise a CI error after removal deadline
+    if deadline is not None and "CI" in os.environ and datetime.now() > deadline:
```
janosh, line 75: maybe better to use `os.getenv("CI)` in case it's set to False (should only ever be set to `True` or be absent but just in case)

```suggestion
    if deadline and os.getenv("CI") and datetime.now() > deadline:
```

janosh (changes requested):

monty/dev.py
DanielYang59: Thanks for reviewing and the suggestions. I decide maybe it's better to take an int tuple instead of `datetime` for maximum user (developer) convenience. Also I noticed the docs need to be updated too.

DanielYang59: I didn't change `if deadline is not None` to `if deadline` here as we didn't implement a strict type check...

From [PEP-8](https://peps.python.org/pep-0008/):
>Comparisons to singletons like None should always be done with is or is not, never the equality operators.
>Also, beware of writing if x when you really mean if x is not None – e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

docs/monty.dev.md
```diff
@@ -26,19 +26,6 @@ with a possible replacement.
...
-**NOTE**: If you are using Python >= 2.7, multiprocessing.cpu_count() already
-provides the number of CPUs. In fact, this is the first method tried.
-The purpose of this function is to cater to old Python versions that
-still exist on many Linux style clusters.
-
-Number of virtual or physical CPUs on this system, i.e.
-user/real as output by time(1) when called with an optimally scaling
-userspace-only program. Return -1 if ncpus cannot be detected. Taken from:
-[http://stackoverflow.com/questions/1006289/how-to-find-out-the-number-of](http://stackoverflow.com/questions/1006289/how-to-find-out-the-number-of)-
-cpus-in-python
```
DanielYang59, line 40: Is the docs automatically generated? It seems this method doesn't exist anymore.

monty/json.py
```diff
@@ -328,7 +328,7 @@ def __get_pydantic_core_schema__(cls, source_type, handler):
         if core_schema is None:
             raise RuntimeError("Pydantic >= 2.0 is required for validation")
 
-        s = core_schema.general_plain_validator_function(cls.validate_monty_v2)
+        s = core_schema.with_info_plain_validator_function(cls.validate_monty_v2)
```
DanielYang59, line 331: Replaced from:
```python
test_json.py::TestJson::test_pydantic_integrations
  /home/yang/Developers/monty/monty/json.py:331: DeprecationWarning: `general_plain_validator_function` is deprecated, use `with_info_plain_validator_function` instead.
    s = core_schema.general_plain_validator_function(cls.validate_monty_v2)
```

docs/monty.dev.md
janosh: yes, the docs are auto-generated from doc strings. possible that this needs to be rerun @shyuep?

DanielYang59: Thanks for the input.

janosh:
yeah, i think a warning is worth considering. but we definitely also need a way to make the warning only show up in the repo's own CI, not in that of downstream packages. [that's why i added](https://github.com/materialsproject/pymatgen/pull/3664/commits/76f04f10f0bd134dcbd857e19182f223d6cc18b9)

```py
os.getenv("GITHUB_REPOSITORY") == "materialsproject/pymatgen"
```

but it would be quite clunky having toe pass in the repo owner and name for every deprecation removal reminder. not sure what the best solution is yet

DanielYang59:
> yeah, i think a warning is worth considering. 

I was being stupid. I realize it's already `DeprecationWarning` not `DeprecationError`.

> but we definitely also need a way to make the warning only show up in the repo's own CI, not in that of downstream packages. 

I just came up with a "not so clean" solution but assume it should work. Maybe just nest a `git config --get remote.origin.url` command inside the decorator during the run and get the code's own repo name like `git@github.com:DanielYang59/monty.git`, and compare it with `GITHUB_REPOSITORY`. 

But I'm a little confused myself which url this command would return if it's imported by a downstream package. What do you think? @janosh

janosh:
using `subprocess` to run `git remote -v` makes the assumption that `git` is installed. but that could be fine. we just have to do nothing if `git` is missing. the overarching principle is not to warn unless we're certain we're running in the origin repo's own CI. so unless all criteria are met, we just do nothing

monty/dev.py
```diff
@@ -42,8 +43,47 @@ def deprecated(
...
+        def _is_in_owner_repo() -> bool:
+            """Check if is running in code owner's repo.
+            Only generate reliable check when `git` is installed and remote name
+            is "origin".
+            """
+
+            try:
+                # Get current running repo
+                result = subprocess.run(
+                    ["git", "config", "--get", "remote.origin.url"],
```
janosh, line 61: if you use `git` with https instead of ssh, the URL returned by `git config remote.origin.url` has this format:

```
https://github.com/materialsproject/pymatgen
```

so the `.lstrip` below won't work as expected. 

```py
os.getenv("GITHUB_REPOSITORY") in owner_repo
```

would be better than

```py
owner_repo == os.getenv("GITHUB_REPOSITORY")
```

janosh (changes requested):

monty/dev.py
DanielYang59: Thanks! I was not aware of this difference. Also there is `GitHub Cli` which seems to depend on the protocol chosen (SSH or HTTPS). What about adding another `lstrip(https://github.com/)`? I feel safer to do a complete match but `in` seems much cleaner (though with some chance of being false position).

janosh: i don't think there's chance of a false positive. `git remote` URLs don't have query params and where else would an unexpected `org/repo` (e.g. `materialsproject/pymatgen`) slip in?

DanielYang59: Well as `monty` is designed to be a general purpose package, I'm afraid there might be repo like `org/repo` and `org/repo-docs` or such. In such case, `org/repo` in `org/repo-docs` (or any other repo with a suffix) would be `True`

janosh: good point! ignore what i said 😅

DanielYang59: I think I would still need a reliable unit test...which is kind of tricky. I totally have no idea how to simulate different CI environment (in code owner's repo or not ), other than manually alter environment variables. I would think about this later.

janosh: yes, manually changing env vars is the right way to test this. you can use the `pytest` [`monkeypatch`](https://docs.pytest.org/en/latest/reference/reference.html#pytest.MonkeyPatch.setenv) fixture of `unittest.patch.dict` to modify env vars in a test. they will be auto-reverted when the test finishes

DanielYang59: Thanks. Would try that soon 😃 .

tests/test_dev.py
```diff
@@ -54,58 +43,99 @@ def func_a(self):
...
         with pytest.warns(DeprecationWarning):
-            assert A().classmethod_b() == "b"
+            assert TestClass_deprecationwarning().classmethod_b() == "b"
+
+    def test_deprecated_deadline(self):
+        @deprecated(deadline=(2000, 1, 1))
+        def func_old():
+            pass
+
+        with warnings.catch_warnings(record=True) as w:
```
janosh, line 96: let's avoid single-letter variable names and call this `warnings` (likewise below)

monty/dev.py
```diff
@@ -29,8 +41,61 @@ def deprecated(replacement=None, message=None, category=FutureWarning):
         Original function, but with a warning to use the updated class.
     """
 
-    def craft_message(old, replacement, message):
+    def _convert_date(date: tuple[int, int, int]) -> datetime:
```
janosh, line 44: don't think we need a helper function for this. this can be inlined

janosh (approved):
very nice tests! this looks ready to go after addressing 2 minor comments. @shyuep can you review and/or merge?

tests/test_dev.py
DanielYang59: Sure! Would change that.

monty/dev.py
DanielYang59: Sorry. This looks like an oversight from my side.

DanielYang59:
Noticed another test failure seemingly related `pydantic` (I guess it's not related to this PR):
```bash
============================================ test session starts =============================================
platform darwin -- Python 3.9.6, pytest-8.0.2, pluggy-1.4.0
rootdir: /Users/yang/developer/monty
plugins: cov-4.1.0, xdist-3.5.0
collected 23 items                                                                                            

tests/test_json.py ....................F..                                                              [100%]

================================================== FAILURES ===================================================
_____________________________________ TestJson.test_pydantic_integrations _____________________________________

value = ForwardRef('LimitedMSONClass | dict')
globalns = {'@py_builtins': <module 'builtins' (built-in)>, '@pytest_ar': <module '_pytest.assertion.rewrite' from '/Users/yang/d...<class 'pydantic.main.BaseModel'>, 'ClassContainingDataFrame': <class 'tests.test_json.ClassContainingDataFrame'>, ...}
localns = {'__abstractmethods__': frozenset(), '__annotations__': {'a': 'LimitedMSONClass | dict'}, '__class_vars__': set(), '__doc__': None, ...}

    def eval_type_backport(
        value: Any, globalns: dict[str, Any] | None = None, localns: dict[str, Any] | None = None
    ) -> Any:
        """Like `typing._eval_type`, but falls back to the `eval_type_backport` package if it's
        installed to let older Python versions use newer typing features.
        Specifically, this transforms `X | Y` into `typing.Union[X, Y]`
        and `list[X]` into `typing.List[X]` etc. (for all the types made generic in PEP 585)
        if the original syntax is not supported in the current Python version.
        """
        try:
>           return typing._eval_type(  # type: ignore
                value, globalns, localns
            )

venv/lib/python3.9/site-packages/pydantic/_internal/_typing_extra.py:240: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/typing.py:290: in _eval_type
    return t._evaluate(globalns, localns, recursive_guard)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = ForwardRef('LimitedMSONClass | dict')
globalns = {'@py_builtins': <module 'builtins' (built-in)>, '@pytest_ar': <module '_pytest.assertion.rewrite' from '/Users/yang/d...<class 'pydantic.main.BaseModel'>, 'ClassContainingDataFrame': <class 'tests.test_json.ClassContainingDataFrame'>, ...}
localns = {'__abstractmethods__': frozenset(), '__annotations__': {'a': 'LimitedMSONClass | dict'}, '__class_vars__': set(), '__doc__': None, ...}
recursive_guard = frozenset()

    def _evaluate(self, globalns, localns, recursive_guard):
        if self.__forward_arg__ in recursive_guard:
            return self
        if not self.__forward_evaluated__ or localns is not globalns:
            if globalns is None and localns is None:
                globalns = localns = {}
            elif globalns is None:
                globalns = localns
            elif localns is None:
                localns = globalns
            type_ =_type_check(
>               eval(self.__forward_code__, globalns, localns),
                "Forward references must evaluate to types.",
                is_argument=self.__forward_is_argument__,
            )
E           TypeError: unsupported operand type(s) for |: 'type' and 'type'

/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/typing.py:546: TypeError

The above exception was the direct cause of the following exception:

self = <tests.test_json.TestJson object at 0x13f882f40>

    def test_pydantic_integrations(self):
        from pydantic import BaseModel, ValidationError
    
        global ModelWithMSONable  # allow model to be deserialized in test
        global LimitedMSONClass
    
        class ModelWithMSONable(BaseModel):
            a: GoodMSONClass
    
        test_object = ModelWithMSONable(a=GoodMSONClass(1, 1, 1))
        test_dict_object = ModelWithMSONable(a=test_object.a.as_dict())
        assert test_dict_object.a.a == test_object.a.a
        dict_no_class = test_object.a.as_dict()
        dict_no_class.pop("@class")
        dict_no_class.pop("@module")
        test_dict_object = ModelWithMSONable(a=dict_no_class)
        assert test_dict_object.a.a == test_object.a.a
    
        assert test_object.model_json_schema() == {
            "title": "ModelWithMSONable",
            "type": "object",
            "properties": {
                "a": {
                    "title": "A",
                    "type": "object",
                    "properties": {
                        "@class": {"enum": ["GoodMSONClass"], "type": "string"},
                        "@module": {"enum": ["tests.test_json"], "type": "string"},
                        "@version": {"type": "string"},
                    },
                    "required": ["@class", "@module"],
                }
            },
            "required": ["a"],
        }
    
        d = jsanitize(test_object, strict=True, enum_values=True, allow_bson=True)
        assert d == {
            "a": {
                "@module": "tests.test_json",
                "@class": "GoodMSONClass",
                "@version": "0.1",
                "a": 1,
                "b": 1,
                "c": 1,
                "d": 1,
                "values": [],
            },
            "@module": "tests.test_json",
            "@class": "ModelWithMSONable",
            "@version": "0.1",
        }
        obj = MontyDecoder().process_decoded(d)
        assert isinstance(obj, BaseModel)
        assert isinstance(obj.a, GoodMSONClass)
        assert obj.a.b == 1
    
        # check that a model that is not validated by pydantic still
        # gets completely deserialized.
        global ModelWithDict  # allow model to be deserialized in test
    
        class ModelWithDict(BaseModel):
            a: dict
    
        test_object_with_dict = ModelWithDict(a={"x": GoodMSONClass(1, 1, 1)})
        d = jsanitize(
            test_object_with_dict, strict=True, enum_values=True, allow_bson=True
        )
        assert d == {
            "a": {
                "x": {
                    "@module": "tests.test_json",
                    "@class": "GoodMSONClass",
                    "@version": "0.1",
                    "a": 1,
                    "b": 1,
                    "c": 1,
                    "d": 1,
                    "values": [],
                }
            },
            "@module": "tests.test_json",
            "@class": "ModelWithDict",
            "@version": "0.1",
        }
        obj = MontyDecoder().process_decoded(d)
        assert isinstance(obj, BaseModel)
        assert isinstance(obj.a["x"], GoodMSONClass)
        assert obj.a["x"].b == 1
    
        # check that if an MSONable object raises an exception during
        # the model validation it is properly handled by pydantic
        global ModelWithUnion  # allow model to be deserialized in test
        global ModelWithLimited  # allow model to be deserialized in test
    
        class ModelWithLimited(BaseModel):
            a: LimitedMSONClass
    
>       class ModelWithUnion(BaseModel):

tests/test_json.py:827: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
venv/lib/python3.9/site-packages/pydantic/_internal/_model_construction.py:178: in __new__
    set_model_fields(cls, bases, config_wrapper, types_namespace)
venv/lib/python3.9/site-packages/pydantic/_internal/_model_construction.py:452: in set_model_fields
    fields, class_vars = collect_model_fields(cls, bases, config_wrapper, types_namespace, typevars_map=typevars_map)
venv/lib/python3.9/site-packages/pydantic/_internal/_fields.py:122: in collect_model_fields
    type_hints = get_cls_type_hints_lenient(cls, types_namespace)
venv/lib/python3.9/site-packages/pydantic/_internal/_typing_extra.py:212: in get_cls_type_hints_lenient
    hints[name] = eval_type_lenient(value, globalns, localns)
venv/lib/python3.9/site-packages/pydantic/_internal/_typing_extra.py:224: in eval_type_lenient
    return eval_type_backport(value, globalns, localns)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

value = ForwardRef('LimitedMSONClass | dict')
globalns = {'@py_builtins': <module 'builtins' (built-in)>, '@pytest_ar': <module '_pytest.assertion.rewrite' from '/Users/yang/d...<class 'pydantic.main.BaseModel'>, 'ClassContainingDataFrame': <class 'tests.test_json.ClassContainingDataFrame'>, ...}
localns = {'__abstractmethods__': frozenset(), '__annotations__': {'a': 'LimitedMSONClass | dict'}, '__class_vars__': set(), '__doc__': None, ...}

    def eval_type_backport(
        value: Any, globalns: dict[str, Any] | None = None, localns: dict[str, Any] | None = None
    ) -> Any:
        """Like `typing._eval_type`, but falls back to the `eval_type_backport` package if it's
        installed to let older Python versions use newer typing features.
        Specifically, this transforms `X | Y` into `typing.Union[X, Y]`
        and `list[X]` into `typing.List[X]` etc. (for all the types made generic in PEP 585)
        if the original syntax is not supported in the current Python version.
        """
        try:
            return typing._eval_type(  # type: ignore
                value, globalns, localns
            )
        except TypeError as e:
            if not (isinstance(value, typing.ForwardRef) and is_backport_fixable_error(e)):
                raise
            try:
                from eval_type_backport import eval_type_backport
            except ImportError:
>               raise TypeError(
                    f'You have a type annotation {value.__forward_arg__!r} '
                    f'which makes use of newer typing features than are supported in your version of Python. '
                    f'To handle this error, you should either remove the use of new syntax '
                    f'or install the `eval_type_backport` package.'
                ) from e
E               TypeError: You have a type annotation 'LimitedMSONClass | dict' which makes use of newer typing features than are supported in your version of Python. To handle this error, you should either remove the use of new syntax or install the `eval_type_backport` package.

venv/lib/python3.9/site-packages/pydantic/_internal/_typing_extra.py:249: TypeError
============================================== warnings summary ===============================================
tests/test_json.py::TestJson::test_pydantic_integrations
tests/test_json.py::TestJson::test_pydantic_integrations
  /Users/yang/developer/monty/venv/lib/python3.9/site-packages/pydantic/main.py:1024: PydanticDeprecatedSince20: The `dict` method is deprecated; use `model_dump` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.6/migration/
    warnings.warn('The `dict` method is deprecated; use `model_dump` instead.', category=PydanticDeprecatedSince20)

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================================== short test summary info ===========================================
FAILED tests/test_json.py::TestJson::test_pydantic_integrations - TypeError: You have a type annotation 'LimitedMSONClass | dict' which makes use of newer typing features t...
================================== 1 failed, 22 passed, 2 warnings in 1.02s ===================================
```

janosh:
not sure why this is popping up now but you need to replace

https://github.com/materialsvirtuallab/monty/blob/1a91828d496df92541e83b9d2d16f6286961d670/tests/test_json.py#L828

with

```py
Union[LimitedMSONClass, dict]
```

DanielYang59:
> not sure why this is popping up now and not before but you need to replace
> 
> https://github.com/materialsvirtuallab/monty/blob/1a91828d496df92541e83b9d2d16f6286961d670/tests/test_json.py#L828
> 
> with
> 
> ```python
> Union[LimitedMSONClass, dict]
> ```

Yes it's odd. Not sure what is happening as this PR didn't touch anything about the Union operator. This operator `|` should be introduced in `Python 3.10`.

And I assume the Python version requirement has to be updated too:
https://github.com/materialsvirtuallab/monty/blob/4d60fd4745f840354e7b3f48346eaf3cd68f2b35/pyproject.toml#L12

And from the main `README`:
https://github.com/materialsvirtuallab/monty/blob/4d60fd4745f840354e7b3f48346eaf3cd68f2b35/README.md?plain=1#L27

This is a bit misleading.

janosh:
This operator | should be introduced in Python 3.10.

it's a `pydantic` restriction which doesn't support forward references below 3.10. see https://github.com/pydantic/pydantic/issues/3300. so you can't use pipe for type unions in any file that defines a `pydantic` schema

tests/test_json.py
```diff
@@ -825,7 +826,7 @@ class ModelWithLimited(BaseModel):
             a: LimitedMSONClass
 
         class ModelWithUnion(BaseModel):
-            a: LimitedMSONClass | dict
+            a: Union[LimitedMSONClass, dict]
```
DanielYang59, line 829: Done.

monty/json.py
DanielYang59: Should fix #598.

janosh:
@shyuep could you review?

janosh:
@shyuep ping

PR #635, opened 2024-03-01
Add optional arg `target_dir` in `compress_file` and `decompress_file` to allow specify target path

### LINKED ISSUE #632: [Feature Request]: `compress_file` and `decompress_file` allow specify target path
### Problem

Would it be a good idea to allow user to optionally specific the output file path for `compress_file` and `decompress_file`?  I could help implement this if so.

Or is there any reason not to do so? 

Thanks

### Proposed Solution

Append a prefix to `filepath` (if not None) should suffice:
https://github.com/materialsvirtuallab/monty/blob/4d60fd4745f840354e7b3f48346eaf3cd68f2b35/monty/shutil.py#L79-L87

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

DanielYang59:
## Summary

- Add optional arg `target_dir` in `compress_file` and `decompress_file` to allow specify target path, to close #632.
- Minor docstring cleanups.

## Summary by CodeRabbit

- **New Features**
	- Introduced an option to specify the target directory for compressing and decompressing files, enhancing user control over file storage locations.

- **Refactor**
	- Removed an unused docstring to improve code clarity.

- **Tests**
	- Added tests for the new target directory feature in file compression and decompression.
	- Improved existing tests for better reliability and coverage.

### DIFF
--- a/monty/io.py
+++ b/monty/io.py
@@ -17,7 +17,7 @@ from typing import IO, Generator, Union
 
 
 def zopen(filename: Union[str, Path], *args, **kwargs) -> IO:
-    r"""
+    """
     This function wraps around the bz2, gzip, lzma, xz and standard python's open
     function to deal intelligently with bzipped, gzipped or standard text
     files.
--- a/monty/shutil.py
+++ b/monty/shutil.py
@@ -6,7 +6,7 @@ import shutil
 import warnings
 from gzip import GzipFile
 from pathlib import Path
-from typing import Literal
+from typing import Literal, Optional
 
 from .io import zopen
 
@@ -66,7 +66,7 @@ def gzip_dir(path: str | Path, compresslevel: int = 6) -> None:
 def compress_file(
     filepath: str | Path,
     compression: Literal["gz", "bz2"] = "gz",
-    target_dir: str | Path = "",
+    target_dir: Optional[str | Path] = None,
 ) -> None:
     """
     Compresses a file with the correct extension. Functions like standard
@@ -78,16 +78,16 @@ def compress_file(
         compression (str): A compression mode. Valid options are "gz" or
             "bz2". Defaults to "gz".
         target_dir (str | Path): An optional target dir where the result compressed
-            file would be stored. Defaults to in-place compression.
+            file would be stored. Defaults to None for in-place compression.
     """
     filepath = Path(filepath)
-    target_dir = str(target_dir)
+    target_dir = Path(target_dir) if target_dir is not None else None
 
     if compression not in {"gz", "bz2"}:
         raise ValueError("Supported compression formats are 'gz' and 'bz2'.")
 
     if filepath.suffix.lower() != f".{compression}" and not filepath.is_symlink():
-        if target_dir:
+        if target_dir is not None:
             os.makedirs(target_dir, exist_ok=True)
             compressed_file = os.path.join(
                 target_dir, f"{os.path.basename(filepath)}.{compression}"
@@ -121,7 +121,9 @@ def compress_dir(path: str | Path, compression: Literal["gz", "bz2"] = "gz") -> 
     return None
 
 
-def decompress_file(filepath: str | Path, target_dir: str | Path = "") -> str | None:
+def decompress_file(
+    filepath: str | Path, target_dir: Optional[str | Path] = None
+) -> str | None:
     """
     Decompresses a file with the correct extension. Automatically detects
     gz, bz2 or z extension.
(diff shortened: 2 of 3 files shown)


monty/shutil.py
```diff
@@ -75,15 +77,25 @@ def compress_file(
...
+        if target_dir:
+            compressed_file = Path(target_dir) / filepath.with_suffix(
+                filepath.suffix + f".{compression}"
+            )
+        else:
+            compressed_file = filepath.with_suffix(filepath.suffix + f".{compression}")
+
+        with open(filepath, "rb") as f_in, zopen(compressed_file, "wb") as f_out:
             f_out.writelines(f_in)
+
```
DanielYang59, line 98: Made this method slightly more sparse for readability

monty/io.py
```diff
@@ -17,7 +17,7 @@
 
 
 def zopen(filename: Union[str, Path], *args, **kwargs) -> IO:
-    r"""
```
DanielYang59, line 20: I guess this is a typo and some accidential remaining from previous docstring changes?

monty/shutil.py
```diff
@@ -75,15 +77,27 @@ def compress_file(
         filepath (str | Path): Path to file.
         compression (str): A compression mode. Valid options are "gz" or
             "bz2". Defaults to "gz".
+        target_dir (str | Path): An optional target dir where the result compressed
+            file would be stored. Defaults to in-place compression.
     """
     filepath = Path(filepath)
-    if compression not in ["gz", "bz2"]:
+    target_dir = str(target_dir)
```
DanielYang59, line 84: Might need a more elegant way here. If using `Path` for `target_dir` directly, the default `if Path("")` would becomes `True`

DanielYang59:
I didn't include `target_dir` for `decompress_dir` suggested by @Andrew-S-Rosen  https://github.com/materialsvirtuallab/monty/issues/632#issuecomment-1972091686 because `decompress_dir` looks more like a convenient tool to reduce the size of a directory to me(`Note that this compresses all files singly, i.e., it does not create a tar archive.`), instead of compressing the entire directory into a single file. So it might be consuing to relocate the compressed files?

monty/shutil.py
```diff
@@ -64,7 +64,9 @@ def gzip_dir(path: str | Path, compresslevel: int = 6) -> None:
 
 
 def compress_file(
-    filepath: str | Path, compression: Literal["gz", "bz2"] = "gz"
+    filepath: str | Path,
+    compression: Literal["gz", "bz2"] = "gz",
+    target_dir: str | Path = "",
```
Andrew-S-Rosen, line 69: I would recommend making the default "." (for current working directory), as "" is not clear.

Same comment in the function below.

monty/shutil.py
```diff
@@ -75,15 +77,28 @@ def compress_file(
         filepath (str | Path): Path to file.
         compression (str): A compression mode. Valid options are "gz" or
             "bz2". Defaults to "gz".
+        target_dir (str | Path): An optional target dir where the result compressed
+            file would be stored. Defaults to in-place compression.
     """
     filepath = Path(filepath)
-    if compression not in ["gz", "bz2"]:
+    target_dir = str(target_dir)
```
Andrew-S-Rosen, line 84: Most of the operations in the module are using the `pathlib` library, so it seems better to convert this to `Path` and use such operations throughout.

Andrew-S-Rosen (commented):
Some initial comments. Afterwards, the same should probably be done to `decompress_dir`.

monty/shutil.py
DanielYang59: If we change the default to ".", the default behaviour would change (from compress files in place to compress files to `cwd`)? Of maybe use `None` as default for in-place compression?

monty/shutil.py
DanielYang59: Yes I find this funny too... But I need to rethink about how to handle and detect the default value if using `Path`

monty/shutil.py
Andrew-S-Rosen: Ah, sorry about that. You're right. The default should be `None`.

monty/shutil.py
Andrew-S-Rosen: This should be self explanatory with a `None` default.

monty/shutil.py
DanielYang59: All good. Would change that.

DanielYang59:
> Can't the same be said for `decompress_file`? 😅 The same suggests that the file is decompressed but doesn't imply it's to be written somewhere else.

Sorry but I still hesitate to add `target_dir` for `(de)compress_dir` as their behaviour is quite different. (As you might know `(de)compress_dir` don't create a single archive file from a `dir`, but perform `(de)compress_file` recursively.)

For `(de)compress_file`, we are basically replace a file with its (de)compressed counterpart, and as such we have full control over the target file.

For `(de)compress_dir`, the behaviour is much more implicit (as it would skip already compressed files): https://github.com/materialsvirtuallab/monty/blob/4d60fd4745f840354e7b3f48346eaf3cd68f2b35/monty/shutil.py#L90-L106

For example, if you `compress_dir` a dir like this to a non-default `target_dir`:
```
source_dir/
├── compressed.gz
├── file1
└── file2
```

The result dir would become:
```
target_dir/
├── file1.gz
└── file2.gz
```

As such this operation becomes irreversible (unless we change the behaviour to either `compress_file` or copy as is, but I personally think this is going too far).

So I think it might be better to leave `(de)compress_dir` as a tool to compress a dir in place?

Andrew-S-Rosen:
That's a fair point! This all makes sense to me.

DanielYang59:
Can you please review this as well? @shyuep

DanielYang59:
Thanks for reviewing @shyuep. The AI reviewer is awesome. Possible to have it for Pymatgen as well? 😄

shyuep:
I just added it to pymatgen. I was giving it a test on simpler repos like monty and matgl first. Pymatgen is a complicated beast.

DanielYang59:
> I just added it to pymatgen. I was giving it a test on simpler repos like monty and matgl first. Pymatgen is a complicated beast.

Great to know. I would keep you updated.

PR #638, opened 2024-03-04
Make `jsanitize(recursive_msonable=True)` respect duck typing

Andrew-S-Rosen:
"If it looks like a duck and quacks like a duck..."

In this PR, I replace the `isinstance(obj, MSONable)` check with a `try`/`except` on the `.as_dict()` method instead in `jsanitize`. This makes it possible to jsanitize objects that have been patched with `.as_dict()` attributes even if they are not formally inheriting from the `MSONable` class.

### DIFF
--- a/monty/itertools.py
+++ b/monty/itertools.py
@@ -1,6 +1,7 @@
 """
 Additional tools for iteration.
 """
+
 import itertools
 
 try:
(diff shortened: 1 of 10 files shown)


orionarcher:
Would love to see this merged, it would unblock my use of duck typing too.

PR #646, opened 2024-03-30
Fix `MSONable.REDIRECT` when module name changed

janosh:
Only class name redirect was actually working. Fixed and now testing both cases.

## Summary by CodeRabbit

- **Refactor**
	- Improved class redirection handling in JSON processing.
- **Tests**
	- Enhanced tests for class redirection in JSON processing, including handling of `ImportError`.
- **Chores**
	- Updated linting workflow to use new `ruff` commands.
	- Updated repository revisions for pre-commit configurations.
- **Chores**
	- Added import statements for various modules in multiple files.
	- Added a docstring to `shutil.py`.

PR #647, opened 2024-03-31
Fix `dev.deprecated` and refine its unit tests

DanielYang59:
### Summary:

- Fix `dev.deprecated` decorator discovered by @janosh in #646
- Refine unit test for `dev.deprecated` to generate more consistent results
- Add `ipython` to `requirements-optional.txt`, [previously getting](https://github.com/materialsvirtuallab/monty/actions/runs/8498187657/job/23277636513):
```
tests/test_dev.py::TestDecorator::test_install_except_hook
  /Users/runner/work/monty/monty/monty/dev.py:185: UserWarning: Cannot install excepthook, IPyhon.core.ultratb not available
    warnings.warn("Cannot install excepthook, IPyhon.core.ultratb not available")
```

## Summary by CodeRabbit

- **New Features**
	- Added `ipython==8.22.2` to optional dependencies for enhanced interactive console features.
- **Bug Fixes**
	- Improved handling of deadlines and deprecation warnings in development tools, ensuring clearer communication to developers.
- **Refactor**
	- Updated test suites for better coverage and clarity in testing deprecation and deadline functionalities.
	- Streamlined code by removing unused imports and unnecessary conditional blocks in test files.


DanielYang59:
My `raise_deadline_warning` addition to `dev.deprecate` seems terribly flawed:
- Accidentally wrong usage of `deadline`(`tuple[int, int, int]`) in string format `{deadline:%Y-%m-%d}`, should be `{_deadline:%Y-%m-%d}`
- Wrong placement of `raise_deadline_warning`, should be inside the `wrapped` function, and therefore the two warnings we noticed previously in CI test.

However even with above two issues fixed, the unit test is still failing in my [forked repo](https://github.com/DanielYang59/monty/actions/runs/8495396336/job/23271639378), where the first "no warn" case ("date before deadline") still triggers a warning, suggesting the time patch in unit test seems unsuccessful (I assume the same warning is not triggered here because this is not "the code owner's repo").  Also the unit tests pass locally. 

I don't quite understand this behavior now, can you please give me a hand on this? @janosh Thanks!

janosh:
> * Wrong placement of `raise_deadline_warning`, should be inside the `wrapped` function, and therefore the two warnings we noticed previously in CI test.

are you sure about that? we want the remove-deprecation error to trigger as soon as the code is parsed by the python interpreter, not wait until the function is called, right?

> However even with above two issues fixed, the unit test is still failing in my [forked repo](https://github.com/DanielYang59/monty/actions/runs/8495396336/job/23271639378), where the first "no warn" case ("date before deadline") still triggers a warning, suggesting the time patch in unit test may be unsuccessful?

i think you may have `monkeypatch`ed incorrectly. you just want to patch `now`, not the `datetime` class, no?


```py
def test_deprecated_deadline_no_warn(self, monkeypatch):
    """Test cases where no warning should be raised."""

    @deprecated(deadline=(2000, 1, 1))
    def func_old():
        pass
    
    monkeypatch.setattr(datetime.datetime, "now", lambda: datetime.datetime(1999, 1, 1)) # changed

    # No warn case 1: date before deadline
    with warnings.catch_warnings(record=True) as warn_msgs:
        func_old()

        for warning in warn_msgs:
            assert "This function should have been removed on" not in str(warning.message)

```

DanielYang59:
Hi @janosh, thanks a lot for your quick response and sharing your suggestions.

> we want the remove-deprecation error to trigger as soon as the code is parsed by the python interpreter, not wait until the function is called, right?

Yes I think this makes sense, but in both cases, existing unit test was inconsistent and need to be adjusted:
```python
    def test_deprecated_deadline(self):
        @deprecated(deadline=(2000, 1, 1))
        def func_old():
            pass

        with warnings.catch_warnings(record=True) as warn_msgs:
            func_old()  # trigger a warning

            assert "will be removed on 2000-01-01" in str(warn_msgs[0].message)
```

Above unit test assert there would be a warning when `func_old()` is called, not when the following is defined:
```python
@deprecated(deadline=(2000, 1, 1))
        def func_old():
            pass
```

> i think you may have monkeypatched incorrectly. you just want to patch now, not the datetime class, no?

I have been struggling with patching time for a while and I wasn't finding much information by searching. If we do:
```python
monkeypatch.setattr(datetime.datetime, "now", datetime.datetime(1999, 1, 1))
```
There would be an error suggesting `now` is immutable:
```
FAILED tests/test_dev.py::TestDecorator::test_deprecated_deadline_no_warn - TypeError: cannot set 'now' attribute of immutable type 'datetime.datetime'
```

One last thing I realized from this issue is that if we just issue a `DeprecationWarning` instead of `raise`, the CI test will pass with warnings, which could be easily neglected (I doubt much people would check the CI run log when CI run passes). Maybe still use a stronger `raise`? What do you think?

tests/test_dev.py
```diff
@@ -89,39 +88,38 @@ def classmethod_b(cls):
...
-            # Trigger a warning.
-            func_old()
-            # Verify message
-            assert "will be removed on 2000-01-01" in str(warn_msgs[0].message)
+
+            @deprecated(deadline=(2000, 1, 1))
+            def func_old():
+                pass
+
+            assert "This function should have been removed" in str(warn_msgs[0].message)
```
DanielYang59, line 97: I would need to patch this as well. Because currently the code is being pushed from my fork, and as such it is not "in code owner" repo, therefore no warning would be issue. And this is [passing in my fork](https://github.com/DanielYang59/monty/actions/runs/8497007313/job/23275096376), could confirm my assumption

janosh:
> Maybe still use a stronger raise? What do you think?

yes! i was actually going to suggest defining a custom error for this:

```py
class ExpiredDeprecationError(RuntimeError): # maybe some other parent class
```

> There would be an error suggesting now is immutable:

interesting, didn't know that. [this SO answer](https://stackoverflow.com/a/69061266) looks promising, have you tried it?

tests/test_dev.py
janosh: that should be easy, no? `monkeypatch.setenv("GITHUB_REPOSITORY", "materialsvirtuallab/monty")`

DanielYang59: > `monkeypatch.setenv("GITHUB_REPOSITORY", "materialsvirtuallab/monty")`

This is a good fix, but I'm thinking if we could make it better:
- Somehow patch `_is_in_owner_repo` to always return `True` (I'm still experimenting on this) such that we don't need to hard code the repo name (`"materialsvirtuallab/monty"`). Even when we hard code the repo name, unit test would still fail in forks.
- Set the `CI` variable as well, such that unit test wouldn't fail locally.

DanielYang59:
> yes! i was actually going to suggest defining a custom error for this:
> 
> ```python
> class ExpiredDeprecationError(RuntimeError):
> ```
> 
> maybe [some other parent class](https://docs.python.org/3/library/exceptions.html#exception-hierarchy) instead of `RuntimeError`

I'm thinking about just `raise DeprecationWarning` to make things simple? All we need is to "make the CI fail" and provide the developers enough clue such that they know some deprecated code should have been removed, in this case a `DeprecationWarning` with proper message should suffice. If the developer for any reason do not want CI to fail, they still could choose not to set the `deadline` arg.

> interesting, didn't know that. [this SO answer](https://stackoverflow.com/a/69061266) looks promising, have you tried it?

Thanks for providing this, I would have a look and let you know.

tests/test_dev.py
DanielYang59: I experimented on patching `_is_in_owner_repo` and did have much luck, then I decide to patch  `subprocess` instead (bcc50019cd101458f0dc211d70a1e1c143b1e700):
```python
with patch("subprocess.run") as mock_run:
                # Mock "GITHUB_REPOSITORY"
                monkeypatch.setenv("GITHUB_REPOSITORY", "TESTOWNER/TESTREPO")
                mock_run.return_value.stdout.decode.return_value = (
                    "git@github.com:TESTOWNER/TESTREPO.git"
                )
```
And everything is working as expected so far.

DanielYang59:
> > There would be an error suggesting now is immutable:
> 
> interesting, didn't know that. [this SO answer](https://stackoverflow.com/a/69061266) looks promising, have you tried it?

Seems adding a `@pytest.fixture()` decorator is all we need to resolve this issue (fadaf1461da1b6d3ba5d5f39a37df582a0471fe9). Meanwhile I further refined its unit test to generate more consistent/reliable results, regardless of the environment.

DanielYang59:
@coderabbitai review

DanielYang59:
Please review this as well @shyuep. Thanks!

DanielYang59:
Thanks for reviewing!

Issue #659, opened 2024-04-16
[Bug]: The `MSONAtoms` class does not rehydrate from `loadfn`


Andrew-S-Rosen:
### Version

b887f02

### Which OS(es) are you using?

- [x] MacOS
- [x] Windows
- [x] Linux

### What happened?

The recently introduced `pymatgen.io.ase.MSONAtoms` class does not properly "rehydrate" from a JSON file via monty's `loadfn` utility even though `MSONAtoms.from_dict()` works appropriately.

#### Setup

```bash
pip install git+https://github.com/materialsvirtuallab/monty.git pymatgen ase
```

#### Example that fails

```python
from monty.serialization import loadfn

loadfn("test.json")
```

[test.json](https://github.com/materialsproject/pymatgen/files/15001117/test.json)

### Log output

```shell
--------------------------------------------------------------------------
TypeError                                Traceback (most recent call last)
Cell In[2], line 1
----> 1 loadfn("test.json")

File c:\Users\asros\miniconda\envs\test\Lib\site-packages\monty\serialization.py:82, in loadfn(fn, fmt, *args, **kwargs)
     80     if "cls" not in kwargs:
     81         kwargs["cls"] = MontyDecoder
---> 82     return json.load(fp, *args, **kwargs)
     84 raise TypeError(f"Invalid format: {fmt}")

File c:\Users\asros\miniconda\envs\test\Lib\json\__init__.py:293, in load(fp, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    274 def load(fp, *, cls=None, object_hook=None, parse_float=None,
    275         parse_int=None, parse_constant=None, object_pairs_hook=None, **kw):
    276     """Deserialize ``fp`` (a ``.read()``-supporting file-like object containing
    277     a JSON document) to a Python object.
    278 
   (...)
    291     kwarg; otherwise ``JSONDecoder`` is used.
    292     """
--> 293     return loads(fp.read(),
    294         cls=cls, object_hook=object_hook,
    295         parse_float=parse_float, parse_int=parse_int,
    296         parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)

File c:\Users\asros\miniconda\envs\test\Lib\json\__init__.py:359, in loads(s, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    357 if parse_constant is not None:
    358     kw['parse_constant'] = parse_constant
[...]

PR #654, opened 2024-04-04
Add save and load functionality to MSONable

matthewcarbone:
This PR adds feature #653 

In particular, it adds `save` and `load` methods to `MSONable`. These methods are flexible, and hook into the existing decoder and encoders in `monty.json`. They work by replacing any instance of an attribute that cannot be serialized by what is essentially a random hash reference to an object in a pickle file saved next to the json. This is then detected during loading, and swapped out for the object when the classmethod `load` is called.

I'd consider this a WIP for now while I continue to test it. I would love feedback by the maintainers @shyuep

## Summary by CodeRabbit

- **New Features**
	- Enhanced serialization and deserialization capabilities with the addition of `save` and `load` methods.
	- Improved handling of unserializable objects in serialization processes.
- **Refactor**
	- Modified the `from_dict` method for improved efficiency and reliability.
- **Tests**
	- Extended testing suite to cover serialization, deserialization, and object equality checks.

Issue #653, opened 2024-04-04
[Feature Request]: pickle fallback for un-MSONable class properties


matthewcarbone:
### Email (Optional)

Preference for just @matthewcarbone in this issue.

### Problem

`monty.json.MSONable` is one of my favorite mixins, and it allows for seamless serialization of classes. However, there are often instances in which classes are unserializable to json due to including attributes that themselves are not serializable, and do not have a defined protocol for serialization in `monty.json`. Obviously, pickling classes directly comes with its drawbacks, but I have an idea.

### Proposed Solution

A new mixin class (we'll call it MSONable2) which defines the `save` and `load` methods. For classes that are completely MSONable, these methods fallback directly on using the json encoder, and saving/loading a json file to the provided path.

However, for classes which are not entirely serializable, for each attribute that isn't, the methods fallback on pickle to save just those objects, one at a time, into a provided directory. This will result in a partially saved json file. I think my idea here is best demonstrated by example.

```python
class TestClass(MSONable2):
	p1 = 123
	p2 = {"hi": "there"}
	p3 = [1, 2, 3, "5"]
	p4 = torch.optimizers.Adam(model.parameters())
```

Above, clearly `p4` is not json-serializable. The way this would work is something like this:

```python
klass = TestClass()
klass.save("path/to/directory/klass.json")
```

Resulting in a json file that looks like this:

```json
{
	"@module": "my.module",
	"@version": "1.2.3",
	"@class": "TestClass",
	"p1": "123",
	"p2": "...",
	"p3": "...",
	"p4": "path/to/directory/klass.pkl",
	"_PICKLE_SERIALIZED_": "p4",
}
```

I'm being lazy with `p2` and `p3`, but you get the idea.

Loading would similarly understand which variables need to be reinstantiated from the pickled data, and which could be loaded normally. Warnings can also be thrown if versions don't match during load.

I can't even begin to explain how useful this would be to me personally, so I imagine others will feel the same way.

### Alternatives

I'm not sure. My alternative is coding it up myself and releasing my own package, but it really feels like this belongs in monty. Happy to do it on my own if you feel it's out of scope though.

Only other alternative is to define more ways of serializing currently unserializable classes.

**I should note I am happy to implement these changes and open a PR.**

@shyuep I see you're making changes to the code recently, what do you think?

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

shyuep:
[...]

PR #669, opened 2024-05-11
Support jsanitize-ing 0D numpy arrays

Andrew-S-Rosen:
Closes #668 by adding support for jsanitize-ing 0D numpy arrays.

PR #670, opened 2024-05-12
Add save and load functionality to MSONable

matthewcarbone:
This PR (re-)adds feature #653.

From #654:

> In particular, it adds `save` and `load` methods to `MSONable`. These methods are flexible, and hook into the existing decoder and encoders in `monty.json`. They work by replacing any instance of an attribute that cannot be serialized by what is essentially a random hash reference to an object in a pickle file saved next to the json. This is then detected during loading, and swapped out for the object when the classmethod `load` is called.

One key difference from the (buggy) #654 is that the decoder was left unchanged. The encoder simply has a fallback when `save` is called, but otherwise behaves identically when the other `MSONable` methods are invoked. 

When `save` is called and an un-`MSONable` object is encountered, it is swapped out for a unique identifier of the form:
```python
{"@object_reference" : unique_id}
```
The object is then stored in a dictionary and pickled next to the saved json.

Any time `"@object_reference"` is encountered during rehydration, it is replaced by the pickled result. I look for these object references by recursively searching the loaded json file (lists and dictionaries only) and replacing these references with their objects before passing to `from_dict`.

> [!IMPORTANT]
> When I tried to implement this before, it came with some breaking bugs due to the fact that MSONable is used in other repositories in ways that weren't completely covered by the testing suite here. Here's an [example](https://github.com/materialsvirtuallab/monty/issues/659#event-12514826617). I was extra careful in this case to not touch any of the default behaviors, and to simply provide fallbacks where exceptions would otherwise be thrown, but that's never a guarantee!

@Andrew-S-Rosen let me know what you think!

### DIFF
--- a/monty/json.py
+++ b/monty/json.py
@@ -472,8 +472,7 @@ class MSONable:
 
     @classmethod
     def load(cls, file_path):
-        """Loads a class from a provided {load_dir}/class.json and
-        {load_dir}/class.pkl file (if necessary).
+        """Loads a class from a provided json file.
 
         Parameters
         ----------
@@ -490,16 +489,24 @@ class MSONable:
         return cls.from_dict(d)
 
 
-def load_anything(path):
-    """Loads a json file into a class, rehydrating from MSONable."""
+def load(path):
+    """Loads a json file that was saved using MSONable.save.
+
+    Parameters
+    ----------
+    path : os.PathLike
+        Path to the json file to load.
+
+    Returns
+    -------
+    MSONable
+    """
 
     d = _d_from_path(path)
     module = d["@module"]
     klass = d["@class"]
-    signature = f"{module}:{klass}"
-    module, function = signature.split(":")
     module = import_module(module)
-    klass = getattr(module, function)
+    klass = getattr(module, klass)
     return klass.from_dict(d)
 
 
--- a/monty/json.py
+++ b/monty/json.py
@@ -8,6 +8,7 @@ import datetime
 import json
 import os
 import pathlib
+import pickle
 import traceback
 import types
 from collections import OrderedDict, defaultdict
@@ -16,7 +17,8 @@ from hashlib import sha1
 from importlib import import_module
 from inspect import getfullargspec
 from pathlib import Path
-from uuid import UUID
+from typing import Any, Dict
+from uuid import UUID, uuid4
 
 try:
     import numpy as np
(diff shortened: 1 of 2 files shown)


Andrew-S-Rosen:
@matthewcarbone: Thanks for revisiting this! And don't worry about the breakage from last time; the tests passed in monty, so you did everything you were supposed to do. You can't be expected to control for things in other repos, but I appreciate your flexibility on this.

This looks pretty safe to me! I did not find any issues with it breaking other repos this time around. 👍

matthewcarbone:
@Andrew-S-Rosen Much appreciated, thanks for checking these things 👍

> You can't be expected to control for things in other repos, but I appreciate your flexibility on this.

Thank you, but still I think I should've been a bit more careful last time. I simply didn't think about the case where people overwrite `from_dict`, but I certainly should have.

Anyways, I'll fix the linting, then we can proceed whenever/if you and @shyuep want to!

matthewcarbone:
@Andrew-S-Rosen it actually looks like the linting error is from a different file: 

```
monty/functools.py:87: error: Argument 1 to "__call__" of "_Wrapper" has incompatible type "lazy_property"; expected "Callable[[VarArg(Never), KwArg(Never)], Never]"  [arg-type]
```

To be completely honest, I'm not _that_ familiar with mypy, but if you'd like I can try to debug this.

Andrew-S-Rosen:
Yeah, it's the reason most of the PRs are currently failing (including mine in #669).

Admittedly, I am also not very familiar with mypy. We will have to wait for @shyuep to review the PRs when he has a moment.

matthewcarbone:
Sounds good. I'll stay in a holding pattern for now. 👍

monty/json.py
```diff
@@ -680,24 +895,12 @@ def jsanitize(
...
     if isinstance(obj, (list, tuple)):
         return [
-            jsanitize(
-                i,
-                strict=strict,
-                allow_bson=allow_bson,
-                enum_values=enum_values,
-                recursive_msonable=recursive_msonable,
-            )
+            jsanitize(i, strict=strict, allow_bson=allow_bson, enum_values=enum_values)
```
Andrew-S-Rosen, line 898: I should have reviewed this better, but the `recursive_msonable` kwarg was removed in a few places. I'll make a patch.

matthewcarbone:
Well that was fast, hoping for no bugs this time 😁

Andrew-S-Rosen:
Pinging @utf, @gpetretto, and @davidwaroquiers since this may be something that impacts jobflow/jobflow-remote development.

PR #671, opened 2024-05-13
Fix internal `jsanitize` calls again

Andrew-S-Rosen:
In #670, the `recursive_msonable` keyword argument was removed from the internal `jsanitize` calls. I have patched it back.


Andrew-S-Rosen:
One more ping from me, @shyuep. 👍 Thanks again!

PR #672, opened 2024-05-13
Add another regression test for `recursive_msonable` behavior

Andrew-S-Rosen:
This PR adds another regression test for `recursive_msonable` behavior in `jsanitize` to prevent breaking of things in the future.

I also ensured the test suite runs if pydantic isn't installed.

### DIFF
--- a/tests/test_json.py
+++ b/tests/test_json.py
@@ -24,6 +24,11 @@ except ImportError:
     torch = None
 
 try:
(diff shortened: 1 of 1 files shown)


Andrew-S-Rosen:
Thanks! I think this should be safe for a new release whenever you feel is appropriate. I have tested pymatgen and a few other repos.

PR #688, opened 2024-06-24
Fix (de)serialization `datetime` with timezone information

### LINKED ISSUE #275: datetime serialization/deserialization broken
datetime objects with time zone info break the serializatin because `__str__` doesn't output in the format that monty expects. Example:
`datetime.datetime(2018, 12, 19, 21, 45, 41, 99000, tzinfo=datetime.timezone.utc).__str__()`
yields
`"2018-12-19 21:45:41.099000+00:00"`

When deserializing, this causes monty to throw an error because the two datetime formatting strings it tries are:
- "%Y-%m-%d %H:%M:%S.%f"
- "%Y-%m-%d %H:%M:%S"

resulting in value error like this from within `strptime`:
`ValueError: unconverted data remains: .967000+00:00`

We could just add the timezone format, or just use a formatting string when outputting from datetime to a string during serialization.  There are also generic datetime parsing libraries like `dateutil`

DanielYang59:
## Summary

- Fix (de)serialization `datetime` with timezone information, to fix #275.
- Replace [deprecated `dict` method with `model_dump`](https://docs.pydantic.dev/2.6/migration/)

#### Alternative one
The following alternative avoids such "hard-coded" timezone info removal, but seems more complicated, not sure which way to go though:
```python
from dateutil import parser


if modname == "datetime" and classname == "datetime":
    try:
        dt = datetime.datetime.strptime(
            parser.parse(d["string"]).strftime(
                "%Y-%m-%d %H:%M:%S.%f"
            ),
            "%Y-%m-%d %H:%M:%S.%f",
        )
    except ValueError:
        dt = datetime.datetime.strptime(
            parser.parse(d["string"]).strftime("%Y-%m-%d %H:%M:%S"),
            "%Y-%m-%d %H:%M:%S",
        )
    return dt

```

#### Alternative two
The following could replace timezone info, but will break if no timezone info is present at all (which thus require a check, might just `split`).

```python
time = str(datetime.datetime.now(datetime.timezone.utc))
print(time)  # 2024-06-27 10:52:45.132087+00:00

dt = datetime.datetime.strptime(time, "%Y-%m-%d %H:%M:%S.%f%z").replace(tzinfo=None)
print(dt)  # 2024-06-27 10:52:45.132087
```

### DIFF
--- a/monty/json.py
+++ b/monty/json.py
@@ -748,6 +748,7 @@ class MontyDecoder(json.JSONDecoder):
                 }:
                     if modname == "datetime" and classname == "datetime":
                         try:
+                            # Remove timezone info in the form of "+xx:00"
                             dt = datetime.datetime.strptime(
                                 d["string"].split("+")[0], "%Y-%m-%d %H:%M:%S.%f"
                             )
--- a/monty/json.py
+++ b/monty/json.py
@@ -766,22 +766,22 @@ class MontyDecoder(json.JSONDecoder):
                     mod = __import__(modname, globals(), locals(), [classname], 0)
                     if hasattr(mod, classname):
                         cls_ = getattr(mod, classname)
-                        d = {k: v for k, v in d.items() if not k.startswith("@")}
+                        data = {k: v for k, v in d.items() if not k.startswith("@")}
                         if hasattr(cls_, "from_dict"):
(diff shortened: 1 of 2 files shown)


DanielYang59:
@shyuep Can you please review this? Thanks.

PR #689, opened 2024-06-24
`deprecated` decorator allow `replacement` as string

DanielYang59:
## Summary

- `deprecated` decorator allow `replacement` as string

Rationale: In some cases, the `replacement` may not be directly accessible from current name space, for example the following where a `property` is marked for deprecation:
```python
class DemoClass:
    def __init__(self, new_property) -> None:
        self.new_property = new_property

    @property
    @deprecated(replacement="new_property")
    def old_property(self):
        return self.new_property
```

### DIFF
--- a/monty/dev.py
+++ b/monty/dev.py
@@ -22,7 +22,7 @@ logger = logging.getLogger(__name__)
 
 
 def deprecated(
-    replacement: Optional[Callable] = None,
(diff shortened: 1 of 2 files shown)


DanielYang59:
@shyuep  I hope this is a rational addition to the `deprecated` decorator. Please review this, thanks!

PR #693, opened 2024-07-01
Add `Path` to supported type hints for `zpath`

Andrew-S-Rosen:
The `monty.os.zpath` function supports `Path`, so I have included this in the type hint (so downstream libraries using `mypy` don't complain).

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the `zpath` function to accept both `str` and `Path` types, improving flexibility.

### DIFF
--- a/monty/os/path.py
+++ b/monty/os/path.py
@@ -11,8 +11,8 @@ from monty.fnmatch import WildCard
 from monty.string import list_strings
 
 if TYPE_CHECKING:
-    from typing import Callable, Literal, Optional, Union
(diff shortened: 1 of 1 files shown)


shyuep:
Seems like the changes broke some type checking.

Andrew-S-Rosen:
Thanks, @shyuep! Missed that. And what do ya know --- `Path` objects aren't actually supported. I must have been looking at the wrong function. This is why we have CI...!

I'll close this and do a proper treatment in a future PR, when I get a spare moment.

PR #699, opened 2024-07-21
Add missing `functools.wraps` decorator to `deprecated` decorator and handle `dataclass` properly

DanielYang59:
### Summary

- Add missing `functools.wraps` decorator to `deprecated` decorator to ensure metadata is passed correctly
- Handle `dataclass` properly, as suggested in https://github.com/materialsproject/atomate2/pull/854#issuecomment-2240262970
- Fix randomly appeared `mypy` errors in `json` module

## Summary by CodeRabbit

- **New Features**
	- Improved handling of deprecated functions and classes for better clarity and usability of warnings, especially for dataclasses.
	- Enhanced type annotation specificity in the JSON handling component, leading to clearer code.

- **Bug Fixes**
	- Ensured proper metadata preservation for deprecated functions and classes to aid in debugging and documentation clarity.

- **Documentation**
	- Updated docstrings to follow Google-style formatting for improved readability and structure.
	- Clarified purposes of deprecated functions and classes through enhanced documentation.

### DIFF
--- a/src/monty/dev.py
+++ b/src/monty/dev.py
@@ -132,8 +132,8 @@ def deprecated(
 
     def deprecated_class_decorator(cls: Type) -> Type:
         # Modify __post_init__ for dataclass
-        if is_dataclass(cls):
-            original_init = cls.__post_init__  # type: ignore[attr-defined]
+        if is_dataclass(cls) and hasattr(cls, "__post_init__"):
+            original_init = cls.__post_init__
         else:
             original_init = cls.__init__
 
@@ -143,8 +143,8 @@ def deprecated(
             warnings.warn(msg, category=category, stacklevel=2)
             original_init(self, *args, **kwargs)
 
-        if is_dataclass(cls):
-            cls.__post_init__ = new_init  # type: ignore[attr-defined]
+        if is_dataclass(cls) and hasattr(cls, "__post_init__"):
+            cls.__post_init__ = new_init
         else:
             cls.__init__ = new_init
 
--- a/src/monty/dev.py
+++ b/src/monty/dev.py
@@ -121,6 +121,7 @@ def deprecated(
         return msg
 
     def deprecated_function_decorator(old: Callable) -> Callable:
+        @functools.wraps(old)
         def wrapped(*args, **kwargs):
             msg = craft_message(old, replacement, message, _deadline)
             warnings.warn(msg, category=category, stacklevel=2)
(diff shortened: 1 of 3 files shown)


src/monty/dev.py
```diff
@@ -129,14 +131,23 @@ def wrapped(*args, **kwargs):
         return wrapped
 
     def deprecated_class_decorator(cls: Type) -> Type:
-        original_init = cls.__init__
+        # Modify __post_init__ for dataclass
+        if is_dataclass(cls) and hasattr(cls, "__post_init__"):
```
janosh, line 135: nice work! 👍 

i think `and hasattr(cls, "__post_init__")` is redundant. all `dataclasses` have `__post_init__`

DanielYang59: Thanks for the comment :)

That was added to prevent `mypy` from complaining:
```
"type[DataclassInstance]" has no attribute "__post_init__"  [attr-defined]
```

Personally I would prefer to avoid `# ignore: ignore` tags wherever we can, what do you think?

janosh: > If no `__init__()` method is generated, then `__post_init__()` will not automatically be called.

not called is different from not present. but if it helps `mypy`, let's keep it but with a comment like `# prevent mypy attr-defined error`

DanielYang59: Thanks for the input. In that case, perhaps we just use the ignore tag to make it self-explanatory.

DanielYang59: It seems there are indeed cases where `__post_init__` doesn't exist:
```python
from dataclasses import dataclass

@dataclass
class SimpleDataClass:
    x: int
    y: int

example_inst = SimpleDataClass(x=10, y=20)
print(hasattr(example_inst, "__post_init__"))  # False
```

janosh: oh my bad, good thing you checked!

DanielYang59: All good. We all make mistakes, **and learn!** :)

DanielYang59:
@shyuep Can you please review this? Not sure why there is a test failure all of a sudden though, doesn't look like related to this PR:

```
 ================================== FAILURES ===================================
________________________________ test_command _________________________________

    def test_command():
        """Test Command class"""
        sleep05 = Command("sleep 0.5")
    
        sleep05.run(timeout=1)
        print(sleep05)
>       assert sleep05.retcode == 0
E       assert 1 == 0
E        +  where 1 = <monty.subprocess.Command object at 0x00000218DC992910>.retcode

tests\test_subprocess.py:12: AssertionError
```

src/monty/json.py
```diff
@@ -17,48 +18,44 @@
...
 try:
     import orjson
 except ImportError:
-    orjson = None  # type: ignore
+    orjson = None
 
-try:
-    import dataclasses
-except ImportError:
-    dataclasses = None  # type: ignore
```
DanielYang59, line 56: I think it's safe to remove guard for `dataclasses` now.
> Added in version 3.7.

PR #700, opened 2024-07-23
Fix line ending in `reverse_readfile/readline` in Windows

### LINKED ISSUE #471: Newlines incorrectly handled in reverse_readfile on windows
## System

* Monty version: 2022.9.9
* Python version: 3.11
* OS version: windows

## Summary

In [reverse_readfile](https://github.com/materialsvirtuallab/monty/blob/23b3c5c390627f773ef04712c022f3f0fdaff61c/monty/io.py#L72) the line separator is hard coded as `\n`, but since monty opens the file in binary mode python doesn't do the usual newline translation you end up with spurious `\r` at the end of lines read by `reverse_readfile`.  I would think `reverse_readlines` suffers from the same problem.  I've came across this only on windows, but a similar issue should happen in macOS, where monty doesn't detect any lines in files, since the line separator is just `\r` there.

## Example code

I don't have a working installation of python+monty on windows, but there's an example output in our CI [here](https://github.com/pyiron/pyiron_base/actions/runs/3884873399/jobs/6628009737#step:9:365).

## Suggested solution (if known)

Just guessing, but a simple solution might just be to open the files in text mode or pass the `newline` argument to the underlying python functions, since you `.decode('utf8')` all strings anyway.  I'm not sure if this would interfere with your handling of compressed files.  If it does you'd have to replace every occurrence of `\n` in the code with `os.linesep`.

DanielYang59:
### Summary

- Fix line ending in `reverse_readfile` and `reverse_readline` in Windows, to fix #471
- Docstring and typing tweaks

## Summary by CodeRabbit

- **New Features**
	- Improved file handling and resource management in the file operations module.
	- Enhanced context management for file locks.

- **Bug Fixes**
	- Updated error handling in YAML import for better type checking.

- **Documentation**
	- Enhanced clarity of docstrings and comments throughout various modules and tests.

- **Refactor**
	- Standardized naming conventions by renaming variables to uppercase constants for improved readability.

- **Tests**
	- Introduced new pytest fixtures to validate line ending handling across platforms.
	- Improved test coverage and maintainability through consistent variable usage and documentation updates.


tests/test_io.py
```diff
@@ -1,15 +1,9 @@
...
 import os
-import unittest
+from pathlib import Path
 
 import pytest
-
-try:
-    from pathlib import Path
-except ImportError:
-    Path = None  # type: ignore
```
DanielYang59, line 11: I think it should be safe to remove this guard against `pathlib` now that we're requiring Python > 3.9. Let me know if there's other reasons that I'm not aware of.
https://github.com/materialsvirtuallab/monty/blob/feb6e8323eb0e074ee5f85838fe67e24682376eb/pyproject.toml#L12

DanielYang59:
@coderabbitai review

DanielYang59:
@shyuep This one is already for review. Thanks!

shyuep:
Thanks

DanielYang59:
Thanks for reviewing!

PR #709, opened 2024-08-22
Drop Python 3.9 support

DanielYang59:
### Summary

- Drop Python 3.9 support
- Bump codecov action to v4


pyproject.toml
```diff
@@ -85,26 +84,24 @@ branch = true
 exclude_also = [
     "@deprecated",
     "def __repr__",
-    "if 0:",
-    "if __name__ == .__main__.:",
```
DanielYang59, line 89: I don't find any of these across the code base, so assume safe to remove

PR #712, opened 2024-09-04
Fix line ending handling in `reverse_readfile/readline` across OS, and not skipping empty lines

DanielYang59:
## I. Fix line ending handling (mainly in Windows with `"\r\n"`)

- [x] Hard coded line ending leads to dangling `\r` for Windows
- [x] Hard coded line ending length `1` prevents `reverse_readline` from working for Windows: https://github.com/materialsvirtuallab/monty/blob/1270c7b230b43cdbc4f5f184ac645ece31d49b61/src/monty/io.py#L131-L133
- [x] [**Behaviour change**] **Yield line as is**, without removing the line ending character (no `rstrip`)

---

## II. Performance benchmark

Compare with current implementation to make sure there's no performance regression, with Python 3.12, using script e4940e04f3d0d01c05dc561db97253b624b4850a.

### Ubuntu 22.04-WSL2

- Installed from PyPI (2024.7.12):  [pypi-7.12-ubuntu2204.txt](https://github.com/user-attachments/files/17043108/pypi-7.12-ubuntu2204.txt)
- Installed from this branch: [develop-ubuntu2204.txt](https://github.com/user-attachments/files/17043112/develop-ubuntu2204.txt)

### Windows 11

- Installed from PyPI ([v2024.7.12](https://github.com/materialsvirtuallab/monty/releases/tag/v2024.7.12)): [pypi-7.12-win11.txt](https://github.com/user-attachments/files/17037632/pypi-7.12-win11.txt)
- Installed from this branch: [develop-win11.txt](https://github.com/user-attachments/files/17037634/develop-win11.txt)

---

## III. Test downstream packages

- `pymatgen`: https://github.com/materialsproject/pymatgen/pull/4068

## Summary by CodeRabbit

- **New Features**
	- Enhanced file handling for compressed files and line endings.
	- Added functionality to determine line endings in files.
  
- **Bug Fixes**
	- Improved error handling for empty files and unknown file types.
  
- **Tests**
	- Introduced new tests for line ending detection and file reading scenarios.
	- Updated existing tests for consistency and clarity across different platforms. 

- **Chores**
	- Minor adjustments to type hints and import statements for clarity.


src/monty/io.py
```diff
@@ -142,17 +159,17 @@ def reverse_readline(
         m_file.seek(0, 2)
         lastchar = m_file.read(1) if is_text else m_file.read(1).decode("utf-8")
 
-        trailing_newline = lastchar == os.linesep
```
DanielYang59, line 145: Looks like `last_char` is not used, we should use it to detect the line ending

hongyi-zhao:
If, for some reason,  the OUTCAR file has mixed style line endings, then how should it be handled? Therefore, I believe that the safest approach might be to first convert the given OUTCAR file to a unified UNIX line ending style.

DanielYang59:
Hi @hongyi-zhao thanks for the comment.

> If, for some reason, the OUTCAR file has mixed style line endings

I believe that's technically possible, but I don't think `reverse_readline/file` should handle cases where files are badly formatted. `reverse_readline/file` would expect the file to be correctly formatted (i.e. has a single line ending across the entire file in our case), if not, the user should correct the format (`dos2unix` or `unit2dos` or such) before sending it into the pipeline.

DanielYang59:
> > `dos2unix` or `unit2dos` or such
> 
> It should be `unix2dos`.

Either serves the same purpose to unify line ending, so it doesn't matter which one you're opting for.

hongyi-zhao:
Don't mind, I mean you have a typo in your previous posting: `unit2dos` should be `unix2dos`.

DanielYang59:
Ah okay, didn't notice that, thanks!

src/monty/io.py
```diff
@@ -54,7 +55,52 @@ def zopen(filename: Union[str, Path], *args, **kwargs) -> IO:
     return open(filename, *args, **kwargs)
 
 
-def reverse_readfile(filename: Union[str, Path]) -> Generator[str, str, None]:
+def _get_line_ending(
```
janosh, line 58: could also make `_get_line_ending` a public function, in which case maybe it belongs in the `monty.os` module.

tests/test_io.py
```diff
@@ -17,6 +19,64 @@
 TEST_DIR = os.path.join(os.path.dirname(__file__), "test_files")
 
 
+class TestGetLineEnding:
```
janosh, line 22: not sure if `_get_line_ending` is meant to handle that but i noticed there's no test for a file with mixed unix/windows line endings which e.g. can happen if you concatenate files from different OSes. one option would be for the function to return a tuple of all the line endings it finds but maybe that's overkill here

janosh (commented):
thanks for fixing this issue @DanielYang59! 👍

tests/test_io.py
DanielYang59: > not sure if `_get_line_ending` is meant to handle that but i noticed there's no test for a file with mixed unix/windows line endings 

@hongyi-zhao https://github.com/materialsvirtuallab/monty/pull/712#issuecomment-2329133801 mentioned this possibility, but I don't think this should be handled by `get_line_ending`. I would expect a user to pass a correctly formatted file into this function (mixed line ending is certainly not a correct format), if not the use should fix the format before passing the file in.

> one option would be for the function to return a tuple of all the line endings it finds but maybe that's overkill here

I'm afraid that would be quite expensive to walk through the file again, perhaps still separate the functionalities, and leave format fixing to another function/tool.

src/monty/io.py
```diff
@@ -186,7 +185,7 @@ def reverse_readline(
     # For windows, we also read the whole file.
     if file_size < max_mem or isinstance(m_file, gzip.GzipFile) or os.name == "nt":
         for line in reversed(m_file.readlines()):
-            yield line.rstrip()
+            yield line.rstrip() if isinstance(line, str) else line.rstrip().decode()
```
DanielYang59, line 188: Might need further confirm (and unit test the type consistency across different OS and file format): https://github.com/materialsvirtuallab/monty/blob/1270c7b230b43cdbc4f5f184ac645ece31d49b61/src/monty/io.py#L89

src/monty/io.py
```diff
@@ -104,68 +157,71 @@ def reverse_readline(
...
         file_size = max_mem + 1
 
-    # If the file size is within our desired RAM use, just reverse it in memory
-    # GZip files must use this method because there is no way to negative seek
+    # If the file size is within desired RAM limit, just reverse it in memory.
+    # GZip files must use this method because there is no way to negative seek.
     # For windows, we also read the whole file.
     if file_size < max_mem or isinstance(m_file, gzip.GzipFile) or os.name == "nt":
         for line in reversed(m_file.readlines()):
-            yield line.rstrip()
```
DanielYang59, line 133: I believe this causes different behavior in Windows and Linux, in Windows (`os.name == "nt"`) the line is right-stripped but not in Linux. 

I think we should NOT preprocess data (`rstrip`) within both `reverse_readfile` and `reverse_readline` because we want it to behave similar to the build in `readlines` just backwards (`readlines` would not `rstrip` lines): https://github.com/materialsvirtuallab/monty/blob/1270c7b230b43cdbc4f5f184ac645ece31d49b61/src/monty/io.py#L114-L116

As such we should treat lines **as is**, just backwards.

src/monty/io.py
```diff
@@ -54,7 +55,53 @@ def zopen(filename: Union[str, Path], *args, **kwargs) -> IO:
...
+
+    Raises:
+        ValueError: If line ending is unknown.
+
+    Warnings:
+        If file is empty, "\n" would be used as default.
+    """
+    if isinstance(file, (str, Path)):
+        with open(file, "rb") as f:
+            first_line = f.readline()
```
DanielYang59, line 78: I think this makes the whole function very memory-expensive, I should rewrite it with `mmap` to read the last several chars instead the entire file.

src/monty/io.py
```diff
@@ -54,7 +55,53 @@ def zopen(filename: Union[str, Path], *args, **kwargs) -> IO:
...
+        return "\r\n"
+    if first_line.endswith(b"\n"):
+        return "\n"
+    if first_line.endswith(b"\r"):
+        return "\r"
+
+    raise ValueError(f"Unknown line ending in line {repr(first_line)}.")
+
+
+def reverse_readfile(
```
DanielYang59, line 102: Current implementation has another issue, say we have a line `repr` as `"line1\nline2\nline3\n"`, when read reversely, it becomes four lines because we just identify lines by finding `\n` (`filemap[line_end_pos + len(l_end) : file_size]`).

src/monty/io.py
```diff
@@ -67,26 +114,32 @@ def reverse_readfile(filename: Union[str, Path]) -> Generator[str, str, None]:
...
+                    yield line.decode("utf-8").rstrip(l_end)
+
             else:
                 filemap = mmap.mmap(file.fileno(), 0, access=mmap.ACCESS_READ)
                 n = len(filemap)
                 while n > 0:
-                    i = filemap.rfind(os.linesep.encode(), 0, n)
-                    yield filemap[i + 1 : n].decode("utf-8").rstrip(os.linesep)
+                    i = filemap.rfind(l_end.encode(), 0, n)
+                    yield filemap[i + 1 : n].decode("utf-8")
```
DanielYang59, line 131: Current implementation has another issue, say we have a line `repr` as `"line1\nline2\nline3\n"`, when read reversely, it becomes four lines because we just identify lines by finding `\n` (`filemap[line_end_pos + len(l_end) : file_size]`).

We should drop the first match.

DanielYang59, line 131: One more question: should we append the line ending when read reversely? (might need to look at how `readline` handles this)

In other word, should we yield the line with or without line ending?

src/monty/io.py
```diff
@@ -54,7 +55,53 @@ def zopen(filename: Union[str, Path], *args, **kwargs) -> IO:
     return open(filename, *args, **kwargs)
 
 
-def reverse_readfile(filename: Union[str, Path]) -> Generator[str, str, None]:
+def _get_line_ending(
```
DanielYang59, line 58: This function has a fundamental weakness, such that I might refrain from using it by default, that is: if the last line doesn't end with a newline character, it would not be possible for it (or any function) to determine its line ending. 

As per [the POSIX standard definition of a line](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206):
> A sequence of zero or more non- <newline> characters plus a terminating <newline> character.

For example if we have the following non-standard file (`repr` as `"line1\nline2"`):
```
line1
line2
```
It would be impossible to find what line ending is used.

As opposed to `"line1\nline2\n"` which would look like (note the "empty line" at the end, technically the last empty line is not a line):
```
line1
line2

```

src/monty/io.py
```diff
@@ -67,26 +119,42 @@ def reverse_readfile(filename: Union[str, Path]) -> Generator[str, str, None]:
...
+                        yield (
+                            filemap[line_end_pos + len(l_end) : file_size].decode(
+                                "utf-8"
+                            )
+                            + l_end
+                        )
+                    count += 1
+                    file_size = line_end_pos
 
     except ValueError:
```
DanielYang59, line 149: Not sure why `except ValueError`, let's see if we could safely remove it.

src/monty/io.py
janosh: sounds like a good idea!

src/monty/io.py
janosh: > It would be impossible to find what line ending is used.

can you elaborate on this? why is it impossible?

DanielYang59: Sorry I didn't intend to make this info obfuscatory (these were written as a note while coding, so might lack detailed explanation). Also **I could be wrong here** because I believe this is already violating the POSIX standard and **discussion on this might not happen / be meaningful in the real world**.

Say if a file complies with POSIX standard (every line has an ending, especially the last), its representation could only ends with two possible the line ending (`"\n"` or `"\r\n"`, [there are more possible line endings](https://en.wikipedia.org/wiki/Newline), but we only consider these two here).

If not, someone could write:
```
line1\r
line2
```
in Linux whose `repr` would be `line1\r\nline2` but we can never know what's it intended to be? Ideally those slash should be skipped as `\\r` but one could also "force-write" them in binary mode.

Say if I write the following files on my mac:
image

With the following code:
```python
with open("file.txt", mode="rb") as file:
    for line in file:
        print(repr(line.decode()))

```
I ended up getting:
```
'line1\\r\r\n'
'line2'
```

I'm already confused with all these crazy cases myself, so perhaps just throw an error if the file is not complying with POSIX standard, instead of trying to guess what's the line ending :)

src/monty/io.py
DanielYang59: I tend not to encourage users to guess line ending through this function (and make it more like a backup plan), because it seems trying to get the line ending from a file is quite prone to all kinds of unexpected errors.

src/monty/io.py
```diff
@@ -67,26 +120,47 @@ def reverse_readfile(filename: Union[str, Path]) -> Generator[str, str, None]:
...
+                            "utf-8"
+                        )
+                    )
+                file_size = seg_start_pos
 
 
 def reverse_readline(
-    m_file, blk_size: int = 4096, max_mem: int = 4000000
-) -> Generator[str, str, None]:
+    m_file,
```
DanielYang59, line 160: A bit unsure why it's `m_file`?

src/monty/io.py
```diff
@@ -104,68 +178,75 @@ def reverse_readline(
...
         file_size = max_mem + 1
 
-    # If the file size is within our desired RAM use, just reverse it in memory
-    # GZip files must use this method because there is no way to negative seek
+    # If the file size is within desired RAM limit, just reverse it in memory.
+    # GZip files must use this method because there is no way to negative seek.
     # For windows, we also read the whole file.
-    if file_size < max_mem or isinstance(m_file, gzip.GzipFile) or os.name == "nt":
+    if (
+        platform.system() == "Windows"
```
DanielYang59, line 208: Not sure why except "windows", look closer back home (with Win PC at home)

src/monty/io.py
```diff
@@ -67,105 +121,135 @@ def reverse_readfile(filename: Union[str, Path]) -> Generator[str, str, None]:
...
+    Read a file backwards line-by-line, and behave similarly to
+    the file.readline function. This allows one to efficiently
+    get data from the end of a file.
 
-    Read file forwards and reverse in memory for files smaller than the
-    max_mem parameter, or for gzip files where reverse seeks are not supported.
+    Cases where file would be read forwards and reversed in RAM:
+    - If file size is smaller than RAM usage limit (max_mem).
+    - In Windows. TODO: explain reason.
+    - For Gzip files, as reverse seeks are not supported.
```
DanielYang59, line 173: It looks like `seek(0, 2)` is supported for gzip now, need performance benchmark before changing implementation.

src/monty/io.py
```diff
@@ -67,105 +121,135 @@ def reverse_readfile(filename: Union[str, Path]) -> Generator[str, str, None]:
...
         buf = ""
         m_file.seek(0, 2)
-        lastchar = m_file.read(1) if is_text else m_file.read(1).decode("utf-8")
+        last_char = m_file.read(1) if is_text else m_file.read(1).decode("utf-8")
 
-        trailing_newline = lastchar == os.linesep
+        trailing_newline = last_char == l_end
 
         while True:
-            newline_pos = buf.rfind(os.linesep)
```
DanielYang59, line 148: Need to rename it because strictly it's not the position of newline, but the end of line

tests/test_io.py
```diff
@@ -138,19 +138,28 @@ def test_reverse_readline(self):
             for idx, line in enumerate(reverse_readline(f)):
                 assert line == f"{str(self.NUMLINES - idx)}{os.linesep}"
 
-    def test_fake_big_file(self):
+    @pytest.mark.parametrize("l_end", ["\n", "\r\n"])
+    def test_big_file(self, l_end):
```
DanielYang59, line 142: This failure is a huge issue (reading big file is the purpose of this function), perhaps how the buffer is refilled or the pointer is advanced has two minor issue (don't know exactly ATM, haven't looked closely yet), leading to off-by-one error in generated line.

Really need to test this thoroughly, and test pymatgen against this before releasing.

DanielYang59:
@shyuep Can I have your advice on this please? 

For a 5 GB file (178_466_370 lines), with current implementation on **Windows** (not really reverse read but read everything with `readlines` first): https://github.com/materialsvirtuallab/monty/blob/1270c7b230b43cdbc4f5f184ac645ece31d49b61/src/monty/io.py#L131-L133

The runtime:
```
Last line 178466370 read, time taken: 34.63317840 s.
75% line 133849777 read, time taken: 41.73865510 s.
50% line 89233185 read, time taken: 50.30206010 s.
```

However if we "fix" the implementation and really read backwards, it get slower (as expected):
```
Last line 178466370 read, time taken: 0.00158920 s.
75% line 133849777 read, time taken: 90.74261630 s.
50% line 89233185 read, time taken: 182.34117580 s.
```

Do we want to change the implementation to really read backwards (behave the same as in Linux/MacOS), or keep it as is? I would personally prefer "fixing" the implementation because: 
- If the user want a "fast" implementation, they can go for `reverse_readfile` (if they want to use filename) or call `readlines()` themselves (if they want to use file stream)?
- Current implementation is not doing what it appears to do (though being faster), which might be unexpected from the user's side.

---

The runtime of `readline` from the start (largely agree with the runtime with current implementation):
```
with open(file_path, "r") as f:
        for _ in range(line_idx + 1):
            line = f.readline()
```

```
Last line 178466370 read, time taken: 40.74100250 s.
75% line 133849777 read, time taken: 30.38311240 s.
50% line 89233185 read, time taken: 20.11819820 s.
```

---

The runtime of `reverse_readfile`:
```
Last line 178466370 read, time taken: 0.00046110 s.
75% line 133849777 read, time taken: 27.03316890 s.
50% line 89233185 read, time taken: 53.99668290 s.
```

shyuep:
Pls fix the merge conflicts.
Thanks.

DanielYang59:
Thanks, I was about to fix these :)

PR #714, opened 2024-10-20
Declare required and optional dependency

DanielYang59:
### Summary

- Add `rualmel.yaml` and `numpy` to required dependencies (as they are used by more than one module), and remove import guard for `ruamel.yaml` and `numpy`
- Declare the rest as optional dependencies, group by module, should make it easier for users to install
- `lzma` is [built in after python 3.3](https://docs.python.org/3/library/lzma.html)

It would then be much easier to install extra dependencies (`pip install monty[json]` or `pip install monty[optional]` to install all extras) without need to dig into the code to find out what is required

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,9 +19,12 @@ jobs:
 
       - name: ruff
         run: |
+          ruff --version
           ruff check .
           ruff format --check .
 
       - name: mypy
         run: |
+          mypy --version
+          rm -rf .mypy_cache
           mypy src
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,12 +19,9 @@ jobs:
 
       - name: ruff
         run: |
-          ruff --version
           ruff check .
           ruff format --check .
 
       - name: mypy
         run: |
-          mypy --version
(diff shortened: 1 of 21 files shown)


pyproject.toml
```diff
@@ -30,7 +30,6 @@ ci = [
   "monty[optional]",
   "pytest>=8",
   "pytest-cov>=4",
-  "types-orjson",
```
DanielYang59: This should be [unnecessary now](https://pypi.org/project/types-orjson/):
> Note: The orjson package includes type annotations or type stubs since version 3.6.1. Please uninstall the types-orjson package if you use this or a newer version.

[orjson 3.6.1 is released Aug 4, 2021](https://pypi.org/project/orjson/3.6.1/)

DanielYang59:
@shyuep Please review this, appreciate it.

.github/workflows/lint.yml
```diff
@@ -19,12 +19,9 @@ jobs:
 
       - name: ruff
         run: |
-          ruff --version
```
DanielYang59, line 22: Printing version doesn't seem necessary as the install stage would show the version:
image

PR #716, opened 2024-10-21
Add unit test for helper function `json._check_type`

DanielYang59:
### Summary

- Add unit test for helper function `json._check_type`, to use it with confident, ping the original author @mturiansky in case you have any comment, thanks
- Use full qualified name `__qualname__` to include the local scope
- Use `isclass` to check class and update comment

## Summary by CodeRabbit

- **New Features**
	- Enhanced type checking functionality for improved accuracy and clarity.
	- Added unit tests for the `_check_type` function, covering various data types including classes, callables, and popular libraries like NumPy, pandas, and PyTorch.

- **Bug Fixes**
	- Improved robustness of the `_check_type` function to correctly handle class objects and subclasses.

- **Documentation**
	- Updated comments in the `_check_type` function for better clarity on its functionality.


src/monty/json.py
```diff
@@ -113,7 +111,9 @@ class B(A):
         mro = type(obj).mro()
     except TypeError:
         return False
-    return any(o.__module__ + "." + o.__name__ == ts for o in mro for ts in type_str)
+    return any(
+        o.__module__ + "." + o.__qualname__ == ts for o in mro for ts in type_str
```
DanielYang59, line 115: I believe using full qualified name to include the local scope might be better? @mturiansky In case I'm wrong here

src/monty/json.py
```diff
@@ -113,7 +111,9 @@ class B(A):
         mro = type(obj).mro()
     except TypeError:
```
DanielYang59, line 112: I could be wrong here, I don't think this `TypeError` is related to callable type:
```python
def func():
    print("hello")

print(type(func).mro())  # >>> [<class 'function'>, <class 'object'>]
```

But more likely that we cannot call `type` on a class constructor:
```python
class A:
    def __init__(self):
        pass

print(type(A).mro())  
```

Which gives:
```
Traceback (most recent call last):
    print(type(A).mro())
          ^^^^^^^^^^^^^
TypeError: unbound method type.mro() needs an argument
```

src/monty/json.py
mturiansky: I don't claim to be an expert on any of this. I just came up with a working solution for the problem I was trying to address. If you think this is a better approach and leads to the same result, then no push-back from me. I'd need to refresh myself on what half of these variables/functions do to be able to make any meaningful comment.

DanielYang59: > I just came up with a working solution for the problem I was trying to address.

Thanks a ton for making everyone's life easier in the first place!

> If you think this is a better approach and leads to the same result

It just covers more edge cases but with unit test, it seems to do what we want it to do. Free feel to comment if I miss any type that should be tested :)

> I'd need to refresh myself on what half of these variables/functions do to be able to make any meaningful comment.

Take your time I would leave this as draft for a while

DanielYang59: I would mark it as ready for review now and feel free to comment :)

mturiansky:
I took a look. I'd say it all looks reasonable to me.

From my understanding, it seems like cases where `__qualname__` is actually needed over `__name__` are rather contrived, but it does seem to be more correct.

DanielYang59:
Thank you!

> From my understanding, it seems like cases where __qualname__ is actually needed over __name__ are rather contrived, but it does seem to be more correct.

Yes I agree, fully qualified name just covers some edge cases (like getting a local func)

PR #722, opened 2024-11-13
Fix custom dict overriding in `collections`

DanielYang59:
### Summary

- Fix custom dict overriding in `collections`
- When trying to modify a frozen dict, raise `TypeError` instead of `KeyError`, in our case, it's "the setter being called on the inappropriate type" instead of "key missing":
    - For example: When trying to modify a `frozenset` it would raise: `TypeError: 'frozenset' object does not support item assignment`:
    - [KeyError: Raised when a mapping (dictionary) key is not found in the set of existing keys.](https://docs.python.org/3/library/exceptions.html#TypeError)
    - [TypeError: Raised when an operation or function is applied to an object of inappropriate type.](https://docs.python.org/3/library/exceptions.html#TypeError)

---

### Custom dicts don't seem to be overridden correctly/fully in `collections`
https://docs.python.org/3/library/stdtypes.html#mapping-types-dict

### `frozendict`
- [x] When trying to add **new** key-value, wrong error message `TypeError: Cannot overwrite existing key: key` 
- [x] `del/pop` method is not overridden, key could be removed

### `FrozenAttrDict`

```python
from monty.collections import FrozenAttrDict

dct = FrozenAttrDict(a=1)
dct["b"] = 2  # >>> KeyError: 'Cannot overwrite existing key: b' (misleading error msg, "b" is not existing key)
```

## Summary by CodeRabbit

- **New Features**
	- Introduced `ControlledDict` class for enhanced control over dictionary operations.
	- Added new tests for `ControlledDict`, `MongoDict`, and `dict2namedtuple`.

- **Bug Fixes**
	- Improved error handling for operations in `ControlledDict`, `frozendict`, and `Namespace`.

- **Documentation**
	- Enhanced docstrings for clarity and usage guidance across various classes.

- **Tests**
	- Refactored test suite to a function-based structure, improving organization and coverage.

### DIFF
--- a/src/monty/collections.py
+++ b/src/monty/collections.py
@@ -101,7 +101,7 @@ class ControlledDict(collections.UserDict, ABC):
 
         super().update(*args, **kwargs)
 
-    def setdefault(self, key, default=None) -> None:
+    def setdefault(self, key, default=None) -> Any:
         """Forbid adding or updating keys based on _allow_add and _allow_update.
 
         Note: if not _allow_update, this method would NOT check whether the
--- a/src/monty/collections.py
+++ b/src/monty/collections.py
@@ -31,7 +31,7 @@ def tree() -> collections.defaultdict:
 class frozendict(dict):
     """
     A dictionary that does not permit changes. The naming
-    violates PEP 8 to be consistent with standard Python's "frozenset" naming.
+    violates PEP 8 to be consistent with the built-in "frozenset" naming.
     """
 
     def __init__(self, *args, **kwargs) -> None:
@@ -88,7 +88,7 @@ class AttrDict(dict):
 
     Examples:
         >>> dct = AttrDict(foo=1, bar=2)
-        >>> assert dct["foo"] == dct.foo
+        >>> assert dct["foo"] is dct.foo
         >>> dct.bar = "hello"
         >>> assert dct.bar == "hello"
     """
--- a/src/monty/collections.py
+++ b/src/monty/collections.py
@@ -79,7 +79,7 @@ class ControlledDict(collections.UserDict, ABC):
 
     # Override add/update operations
     def __setitem__(self, key, value):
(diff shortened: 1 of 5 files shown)


src/monty/collections.py
```diff
@@ -217,11 +216,11 @@ def dict2namedtuple(*args, **kwargs) -> tuple:
...
     Examples:
-        >>> t = dict2namedtuple(foo=1, bar="hello")
-        >>> assert t.foo == 1 and t.bar == "hello"
+        >>> tpl = dict2namedtuple(foo=1, bar="hello")
+        >>> assert tpl.foo == 1 and tpl.bar == "hello"
 
-        >>> t = dict2namedtuple([("foo", 1), ("bar", "hello")])
-        >>> assert t[0] == t.foo and t[1] == t.bar
+        >>> tpl = dict2namedtuple([("foo", 1), ("bar", "hello")])
+        >>> assert tpl[0] is tpl.foo and t[1] is tpl.bar
```
DanielYang59, line 223: I believe we should check identity instead of value equality in this case, as both should point to exactly the same object

src/monty/collections.py
```diff
@@ -20,126 +30,202 @@ def tree() -> collections.defaultdict:
...
 
+    # Override add/update operations
+    def __setitem__(self, key, value) -> None:
+        """Forbid adding or updating keys based on _allow_add and _allow_update."""
+        if key not in self.data and not self._allow_add:
+            raise TypeError(f"Cannot add new key {key!r}, because add is disabled.")
+        elif key in self.data and not self._allow_update:
+            raise TypeError(f"Cannot update key {key!r}, because update is disabled.")
 
-class Namespace(dict):
```
DanielYang59, line 57: If I understand correctly, `Namespace` is a dict that only allows adding new items but not allow updating existing items, however I don't really understand why it's called `Namespace`?

PR #725, opened 2024-11-27
Add case-insensitive dictionary `CaseInsensitiveDict[Base/Upper/Lower]` to `collections`

DanielYang59:
### Summary

- Add case-insensitive dictionary to `collections`, migrate `pymatgen` implementation of `Lobsterin` and `Incar`

### Performance Benchmark

**Init a `CaseInsensitiveDict` would be way more expensive than a standard dict** (a bare `UserDict` is less performant than standard dict), but lookup is almost instant (typical of a hash table).

Bare `UserDict` vs standard dict:
```
  Size    Dict Init Time (ms)    Dict Lookup Time (ms)    CI Dict Init Time (ms)    CI Dict Lookup Time (ms)
------  ---------------------  -----------------------  ------------------------  --------------------------
     1                  0.000                    0.000                     0.006                       0.000
    10                  0.000                    0.000                     0.002                       0.000
   100                  0.000                    0.000                     0.007                       0.000
  1000                  0.003                    0.000                     0.060                       0.000
 10000                  0.030                    0.000                     0.622                       0.000
100000                  0.442                    0.000                     7.830                       0.001
```

`CaseInsensitiveDictUpper` vs standard dict:
```
  Size    Dict Init Time (ms)    Dict Lookup Time (ms)    CI Dict Init Time (ms)    CI Dict Lookup Time (ms)
------  ---------------------  -----------------------  ------------------------  --------------------------
     1                  0.000                    0.000                     0.008                       0.001
    10                  0.000                    0.000                     0.003                       0.001
   100                  0.001                    0.000                     0.016                       0.000
  1000                  0.003                    0.000                     0.138                       0.000
 10000                  0.019                    0.000                     1.403                       0.001
100000                  0.511                    0.001                    16.021                       0.002
```

---

<details>

<summary> Test Script (by GPT) </summary>

```python
import time
import random
import string
from monty.collections import CaseInsensitiveDictUpper
from tabulate import tabulate


def measure_time_for_initialization(dict_type, size: int) -> float:
    """Measure time to initialize a dictionary of size `size`."""
    data = {random_string(): random.randint(1, 100) for _ in range(size)}
    start_time = time.perf_counter()
    dict_type(data)
    return (time.perf_counter() - start_time) * 1000  # Convert to milliseconds


def measure_time_for_lookup(dict_obj, size: int) -> float:
    """Measure time to look up a random key in the dictionary."""
    key = random.choice(list(dict_obj.keys()))
    start_time = time.perf_counter()
    dict_obj[key]
    return (time.perf_counter() - start_time) * 1000  # Convert to milliseconds


def random_string(length: int = 5) -> str:
    """Generate a random string of given length."""
    return ''.join(random.choices(string.ascii_lowercase, k=length))


def compare_performance():
    sizes = [1, 10, 100, 1000, 10_000, 100_000]
    results = []

    # Compare performance for standard dict and CaseInsensitiveDictUpper
    for size in sizes:
        # Initialize and measure performance for dict
        dict_init_time = measure_time_for_initialization(dict, size)
        dict_lookup_time = measure_time_for_lookup(dict({random_string(): random.randint(1, 100) for _ in range(size)}), size)

        # Initialize and measure performance for CaseInsensitiveDictUpper
        cid_init_time = measure_time_for_initialization(CaseInsensitiveDictUpper, size)
        cid_lookup_time = measure_time_for_lookup(CaseInsensitiveDictUpper({random_string(): random.randint(1, 100) for _ in range(size)}), size)

        results.append([
            size,
            dict_init_time,
            dict_lookup_time,
            cid_init_time,
            cid_lookup_time,
        ])

    # Print the results in a table format
    headers = ["Size", "Dict Init Time (ms)", "Dict Lookup Time (ms)", "CI Dict Init Time (ms)", "CI Dict Lookup Time (ms)"]
    print(tabulate(results, headers=headers, floatfmt=".3f"))


if __name__ == "__main__":
    compare_performance()
```
</details>

### DIFF
--- a/src/monty/collections.py
+++ b/src/monty/collections.py
@@ -243,6 +243,12 @@ class CaseInsensitiveDictUpper(CaseInsensitiveDictBase):
         return key.upper() if isinstance(key, str) else key
 
 
+class CaseInsensitiveDictLower(CaseInsensitiveDictBase):
+    @staticmethod
+    def _converter(key: Any) -> Any:
+        return key.lower() if isinstance(key, str) else key
+
+
 class MongoDict:
     """
     This dict-like object allows one to access the entries in a nested dict as
--- a/src/monty/collections.py
+++ b/src/monty/collections.py
@@ -5,7 +5,8 @@ Useful collection classes, e.g., tree, frozendict, etc.
 from __future__ import annotations
 
 import collections
-from typing import TYPE_CHECKING
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING, Mapping
 
 if TYPE_CHECKING:
     from typing import Any, Iterable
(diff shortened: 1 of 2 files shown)


shyuep:
I am not sure this is needed. It is a very esoteric use case. I'd rather not violate Python's case sensitivity just for those use cases. Those use cases are already handled.

PR #729, opened 2024-12-06
Add support for extended JSON to MontyDecoder

esoteric-ephemera:
When `bson` is found, allows `MontyDecoder` to parse [extended JSON](https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/) values from BSON. This helps with retrieving data which is `float("inf")`, `float("NaN")` as well as `datetime`s without adding custom logic

### DIFF
--- a/src/monty/json.py
+++ b/src/monty/json.py
@@ -29,8 +29,10 @@ if TYPE_CHECKING:
 
 try:
     import bson
+    from bson import json_util
 except ImportError:
     bson = None
+    json_util = None
 
 try:
     import orjson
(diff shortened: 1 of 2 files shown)


tschaume:
@shyuep Merging this PR and publishing a release candidate for monty would be a great help to get MP's newest data release out. Is there anything I can do to help? I'd be happy to take on a maintainer role (for release candidates only) if needed. Monty is used constantly in the python API client, during the storage of our intermediary and final data products in OpenData repositories, and in our web service. We keep monty dependencies up to date in our stack, so potential issues should be on our radar immediately when we use new releases in production.

shyuep:
Release is done. Thanks for the offer to help maintain the project. Currently, I prefer to limit the number of maintainers to a small group. Please let me know if releasing the project becomes an issue. Thanks.

tschaume:
Perfect! Thanks for the quick turnaround!

PR #730, opened 2024-12-07
`zopen` changes: forbid implicit binary/text `mode`, signature change, default UTF-8 `encoding` in text mode, drop `.z` support after one-year

DanielYang59:
### Summary
- Use UTF-8 `encoding` as default in text mode when `encoding` is not given, follow up https://github.com/materialsproject/pymatgen/pull/4218, to fix https://github.com/materialsproject/pymatgen/issues/4214
- Forbid implicit "binary/text" `mode`, **the plan is giving a half-year grace time (I prefer to enforce a shorter grace time as this is not deprecation but more a fix for a dangerous behaviour), after that passing a `mode` without `b/t` would directly trigger an exception**
- `zopen` signature change 
- Test enhance
- [x] Test downstream `pymatgen`: [commit](https://github.com/materialsproject/pymatgen/pull/4219/commits/c9bf970928e7e1d235a835e3cb4ad45a149ce010),  [CI log](https://github.com/materialsproject/pymatgen/actions/runs/12222964584/job/34093905469?pr=4219)

---

### Signature change for `zopen`

- Force `file/filename` as positional, AFAIK, there is [no usage as `with zopen(filename=)` ](https://sourcegraph.com/search?q=context:global+with+zopen%28filename%3D&patternType=keyword&sm=0) (i.e. `filename` as kwarg). 
  - **Rationale**: Different file var names in `open` (`file`) and `gzip/bz2/lzma.open` (`filename`)
- Make arguments other than `filename/mode` keyword only. 
  - **Rationale**: Completely different signature after `filename` and `mode`, passing as positional args would give very unpredictable results:
    - [open(file, mode='r', buffering=-1, encoding=None, errors=None, newline=None, closefd=True, opener=None)](https://docs.python.org/3/library/functions.html#open)
    - [bz2.open(filename, mode='rb', compresslevel=9, encoding=None, errors=None, newline=None)](https://docs.python.org/3/library/bz2.html#bz2.open)
    - [gzip.open(filename, mode='rb', compresslevel=9, encoding=None, errors=None, newline=None)](https://docs.python.org/3/library/gzip.html#gzip.open)
    - [lzma.open(filename, mode='rb', *, format=None, check=-1, preset=None, filters=None, encoding=None, errors=None, newline=None)](https://docs.python.org/3/library/lzma.html#lzma.open)
  
---

### Forbid implicit binary/text mode for `zopen`
- [x] Emit warning when `mode` is not set, or `b/t` is not given in `mode`

**Rationale:**
Using implicit binary/text `mode` (no `b/t` in `mode`) with `zopen` would cause different behaviour for text file (`file.txt`, as `TextIO`) vs compressed file (`file.gz`, as `BinaryIO`), as default mode for `open` is "text" while default mode for compressed `open` is binary, as such we would get different data types from the stream (string vs bytes).

A `read` example:
```python
with zopen("hello.txt") as file:
    print(file.readline())  # "hello" as string

with zopen("hello.txt.gz") as file:
    print((first_line := file.readline()))  # b'hello\n' as bytes
    print("hello" in first_line)  # TypeError: a bytes-like object is required, not 'str'
```

A `write` example:

```python
with zopen("file.gz", mode="w") as file:
    file.write("hello")  # TypeError: memoryview: a bytes-like object is required, not 'str'
```

---

### Default "UTF-8" `encoding` in text mode

Rationale see https://github.com/materialsproject/pymatgen/pull/4218#issue-2722409653

---

### Drop support for `.z` extension after one-year grace period

I don't think Python's built-in `gzip` could (de)compress a [**real LZW-algorithm file**](https://en.wikipedia.org/wiki/Gzip) (not a file (de)compressed by [`gzip` DEFLATE algorithm](https://docs.python.org/3/library/zlib.html#zlib.compressobj) and named to `.z`):
> The gzip file format is also not to be confused with that of the compress utility, based on LZW, with extension .Z

> method is the compression algorithm. Currently, the only supported value is DEFLATED.

For example:
```python
import gzip
import os
import subprocess


with open("hello.txt", mode="wt", encoding="utf-8") as file:
    file.write("hello\n")

# On Ubuntu: `sudo apt install ncompress`, see: https://vapier.github.io/ncompress/
subprocess.run(["compress", "-f", "hello.txt"], check=True)
assert os.path.isfile("hello.txt.Z")

with gzip.open("hello.txt.Z", mode="rt") as file:
    print(file.readlines())  # gzip.BadGzipFile: Not a gzipped file (b'\x1f\x9d')
```

However user might have already used something like `with zopen("text.z", mode="wt") as file:` to generate such a file (looks like LZW but actually DEFLATE), and as such we wouldn't drop `.z` extension directly here, a grace period of one-year would be added.


src/monty/io.py
```diff
@@ -23,30 +23,30 @@
...
-        File-like object. Supports with context.
+        File-like object. Supports context manager `with`.
     """
     if filename is not None and isinstance(filename, Path):
         filename = str(filename)
 
     _name, ext = os.path.splitext(filename)
-    ext = ext.upper()
-    if ext == ".BZ2":
+    ext = ext.lower()
```
DanielYang59, line 41: I might convert file extension to lower instead as I assume this is more conventional (and no behaviour change at all)? I haven't found a naming standard though.

```bash
touch test
gzip test
ls
>>> test.gz 
```

src/monty/io.py
```diff
@@ -21,33 +21,42 @@
...
+        filename (str | Path): Filename.
+        # TODO: mode use Literal type
+        mode (str):  E.g., "r" for read, "w" for write.  # TODO: finish this
+        **kwargs: Keyword arguments to pass to `open`.
 
     Returns:
-        File-like object. Supports with context.
+        TextIO | BinaryIO
     """
-    if filename is not None and isinstance(filename, Path):
```
DanielYang59, line 39: I would drop the `filename` as `None` check and merge `Path` conversion as I don't think opening `None` is a legal usage

src/monty/io.py
```diff
@@ -14,40 +14,89 @@
...
+    # TODO: replace warning with exception after deadline
+    elif not ("b" in mode or "t" in mode):
+        warnings.warn(
+            "We strongly discourage using implicit binary/text `mode`, "
+            f"and this would not be allowed after {_deadline}. "
+            "I.e. you should pass t/b in `mode`, we would assume text mode for now",
+            FutureWarning,
+            stacklevel=2,
+        )  # TODO: unit test
+        mode += "t"  # assume text mode if not specified
```
DanielYang59, line 77: I would assume `mode` to be text here instead of leave it as is for two reasons:
- For `pymatgen`, [most (if not all) implicit binary/text usage seem to assume text mode](https://github.com/search?q=repo%3Amaterialsproject%2Fpymatgen%20with%20zopen&type=code), i.e. apply string-specific operations on the stream instead of bytes
- Facilitate `encoding` check (otherwise cannot use `if "t" in encoding` to determine if text mode is expected)

src/monty/io.py
```diff
@@ -14,40 +14,89 @@
...
+        TextIO | BinaryIO
     """
-    if filename is not None and isinstance(filename, Path):
-        filename = str(filename)
-
-    _name, ext = os.path.splitext(filename)
-    ext = ext.upper()
-    if ext == ".BZ2":
-        return bz2.open(filename, *args, **kwargs)
-    if ext in {".GZ", ".Z"}:
```
DanielYang59, line 46: I don't think Python's built-in `gzip` could open a [**real LZW compressed file**](https://en.wikipedia.org/wiki/Gzip) (not a file compressed by `gzip` DEFLATE algorithm and name to `.z`):
> The gzip file format is also not to be confused with that of the compress utility, based on LZW, with extension .Z

For example:
```python
import gzip
import os
import subprocess


with open("hello.txt", mode="wt", encoding="utf-8") as file:
    file.write("hello\n")

subprocess.run(["compress", "-f", "hello.txt"], check=True)
assert os.path.isfile("hello.txt.Z")

with gzip.open("hello.txt.Z", mode="rt") as file:
    print(file.readlines())  # gzip.BadGzipFile: Not a gzipped file (b'\x1f\x9d')
```

DanielYang59:
Tested on `pymatgen` and there's only one CI failure in https://github.com/materialsproject/pymatgen/blob/31f1e1fb8cc1bb517d59ea8e484965bb641c42a0/tests/io/vasp/test_inputs.py#L1563


I personally would prefer to discourage such usage to pass a `mode` with binary/text to consistently get binary or text, instead of dependent on the file extension (unlike the case of using `gzip.open` where people would expect a BinaryIO).

I have updated that line to:
```diff
- fe_potcar = zopen(f"{FAKE_POTCAR_DIR}/POT_GGA_PAW_PBE/POTCAR.Fe_pv.gz").read().decode("utf-8")
+ fe_potcar = zopen(f"{FAKE_POTCAR_DIR}/POT_GGA_PAW_PBE/POTCAR.Fe_pv.gz", mode="rt", encoding="utf-8").read()
```

src/monty/io.py
```diff
@@ -14,40 +14,104 @@
...
+    if ext == ".z":
+        # TODO: drop ".z" extension support after 2026-01-01
+        warnings.warn(
+            "Python gzip is not able to (de)compress LZW-compressed files. "
+            "You should rename it to .gz. Support for the '.z' extension will "
+            "be removed after 2026-01-01.",
+            category=FutureWarning,
+            stacklevel=2,
+        )
+    if ext in {".xz", ".lzma"}:
```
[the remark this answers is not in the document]
DanielYang59: Big overlook from my side

src/monty/io.py
```diff
@@ -14,40 +14,102 @@
...
+    if "t" in mode and kwargs.get("encoding", None) is None:
+        warnings.warn(
+            "We strongly encourage explicit `encoding`, "
+            "and we would use UTF-8 by default as per PEP 686",
+            category=EncodingWarning,
+            stacklevel=2,
+        )
+        kwargs["encoding"] = "utf-8"
 
     _name, ext = os.path.splitext(filename)
```
DanielYang59, line 92: I guess `splittext` got updated at some point? It works directly with `str` as path now

PR #734, opened 2024-12-28
allow `strict="skip"` for `jsanitize`

jmmshn:
# Optional Alternative Behavior for `jsanitize`

Currently, if we use `strict=False` all of the unserializable objects will be replaced with the `str` representation which will lose information.  The situation I'm thinking of is the case of processing an object that is not necessarily top-level msonable but we still want to convert it to a dictionary first for processing with the `_get_partial_json` mechanism.

This PR adds a new way for `jsanitize` to behave:

If you provide `strict="skip"` it will just ignore any unserializable objects and leave it alone. without converting into a string.

## Summary by CodeRabbit

- **New Features**
	- Enhanced JSON sanitization function with more flexible error handling.
	- Added support for skipping unserializable objects during JSON processing.

- **Tests**
	- Introduced new test case for handling composite objects with mixed serialization capabilities.

- **Bug Fixes**
	- Improved handling of objects that cannot be easily serialized.

### DIFF
--- a/src/monty/json.py
+++ b/src/monty/json.py
@@ -905,7 +905,8 @@ def jsanitize(
             jsanitize will try to get the as_dict() attribute of the object. If
             no such attribute is found, an attribute error will be thrown. If
             strict is False, jsanitize will simply call str(object) to convert
-            the object to a string representation.
+            the object to a string representation.  If "skip" is provided,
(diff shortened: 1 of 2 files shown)


jmmshn:
@coderabbit, can you apply the suggested changes (Enhance test clarity with descriptive messages and complete comments)?

PR #739, opened 2025-01-09
Re-add python 3.9 support

DanielYang59:
## Summary

Re-add python 3.9 support, as the oversight in `dev.deprecated` warning seem to broadly impact downstream packages on Python 3.9 and would require hacky removal of `CI` environment variable 

Also `monty` doesn't really have too many Python version specific functionalities, supporting a broader Python version might bring other fixes to lower Python versions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
       max-parallel: 20
       matrix:
         os: [ubuntu-latest, macos-14, windows-latest]
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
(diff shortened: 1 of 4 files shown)


DanielYang59:
@shyuep Can we re-consider adding back Python 3.9 support? As:
- `monty` doesn't really have too many Python version specific functionality AFAIK, so adding back Python 3.9 would not require any code change AFAIK.
- Python 3.9 is not EOL yet, and as `monty` more is a supplementary package of Python instead of NumPy, perhaps let `monty` follow Python life-cycle while `pymatgen` still follows NEP 29?

shyuep:
Ok with me.

DanielYang59:
Thanks!

PR #746, opened 2025-02-25
Add support for `bson.dbref.DBRef`

mkhorton:
## Summary

`DBRef` is a bson type which holds a pointer to a specific document in a MongoDB collection. This PR adds support for serializing and de-serializing `DBRef` objects to `monty`.

## Summary by CodeRabbit

- **New Features**
	- Enhanced JSON processing to support BSON reference objects, improving the handling of database references in data exchange.
- **Tests**
	- Introduced tests to verify the correct serialization and deserialization of BSON reference objects across various scenarios.
- **Documentation**
	- Updated documentation to outline the new support for BSON reference objects.


mkhorton:
Documentation on `DBRef` is available here:

https://www.mongodb.com/docs/manual/reference/database-references/#format

and in `pymongo` here:

https://pymongo.readthedocs.io/en/stable/api/bson/dbref.html

shyuep:
**🤖 Automated PR review (generated by Claude on behalf of @shyuep)**

- ✅ Symmetric `DBRef` encode/decode mirroring the existing `ObjectId` handling is the right shape, and `o.as_doc()` is the canonical way to dump a `DBRef`.
- ⚠️ **Bug in the new test** (`tests/test_json.py::test_dbref`): `x = json.loads(dbref, cls=MontyDecoder)` is decoding the `DBRef` object itself, not the JSON string `djson` produced above. Should be `json.loads(djson, cls=MontyDecoder)`. As written the test would fail (or trivially pass on `TypeError`) and the round-trip assertions never exercise the decode path.
- ⚠️ CI is red: all 12 `build` matrix jobs fail on the head commit (`d34ad50`). Likely the test bug above; please re-run after fixing.
- ⚠️ **Import block** in `tests/test_json.py` is a bit tangled — the `from bson import json_util` line lives inside the `except ImportError` clause for `DBRef`, so if `DBRef` imports succeed but `json_util` import doesn't, `DBRef` will still get reset to `None`. Split into two independent `try/except` blocks.
- 💡 `branch is dirty` (conflicts with `master`); needs a rebase.
- 💡 Consider asserting `x.id == 1` in the test too to confirm the `$id` field round-trips, and a test where `database` is `None` to confirm the `d.get("$db")` fallback works.
- 💡 Still marked as **draft** — once the test bug and CI are sorted, ready to un-draft.

---

shyuep:
Superseded by #792, which adds bson.dbref.DBRef serialization via the new TypeHandler plugin system (and merges the bson handlers so ObjectId and DBRef share a single class).

A few differences from this PR:

- Wire format uses collection / id / optional database / optional extras instead of bson's $ref / $id / $db. The latter would be intercepted by bson.json_util.loads and reconstructed into a DBRef before our decoder could route the nested `id` field through MontyDecoder.
- The nested id (and any extras) recurse through MontyDecoder().process_decoded, so a DBRef carrying an ObjectId round-trips with id reconstructed as an ObjectId rather than a raw dict.
- jsanitize(allow_bson=True) passes DBRef through unchanged.

Thanks for the original work!

PR #752, opened 2025-05-30
`zopen` wouldn't allow default `mode` nor implicit text/binary `mode`, allow `mode` as positional argument

DanielYang59:
### Summary

Follow up https://github.com/materialyzeai/monty/pull/730

- `io.zopen` wouldn't allow default `mode`, nor implicit text/binary `mode`
- `io.zopen` allows `mode` as positional argument


src/monty/io.py
```diff
@@ -27,8 +27,7 @@ class EncodingWarning(Warning): ...  # Added in Python 3.10
 
 def zopen(
     filename: Union[str, Path],
-    /,
```
DanielYang59, line 30: I guess it's better that we also allow `mode` as a positional argument now that all monty-supported file openers take `mode` as second argument:
https://github.com/materialsvirtuallab/monty/blob/0dd2588667e4bfa0a8436b282b9e1f371ad5abc6/src/monty/io.py#L99-L116

pyproject.toml
```diff
@@ -97,6 +97,9 @@ select = [
 required-imports = ["from __future__ import annotations"]
 known-first-party = ["monty"]
 
+[tool.uv]
+default-groups = "all"
```
DanielYang59, line 101: https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups
> By default, uv includes the dev dependency group in the environment (e.g., during uv run or uv sync). The default groups to include can be changed using the tool.uv.default-groups setting.

esoteric-ephemera:
Hey @DanielYang59 this is a pretty big breaking change for `monty`. We have a lot of legacy code of the form `zopen(file_name)` with an implicit "r" mode, consistent with standard python's `open`

I get that there's a warning for the transition, but is there a good reason not to follow stdlib's default read mode / utf-8 encoding?

DanielYang59:
> but is there a good reason not to follow stdlib's default read mode / utf-8 encoding?

Yes I believe there is (sorry it seems I forgot to link to the original PR https://github.com/materialyzeai/monty/pull/730 here), you could check more details there, quote:
> Using implicit binary/text mode (no b/t in mode) with zopen would cause different behaviour for text file (file.txt, as TextIO) vs compressed file (file.gz, as BinaryIO), as default mode for open is "text" while default mode for compressed open is binary, as such we would get different data types from the stream (string vs bytes).

esoteric-ephemera:
That's fair, but a better middle path solution here may just be normalizing behaviors across different protocols, e.g., defaulting to `"rt"` / `"utf-8"`, or defaulting to the mode/encoding used by a specific `open` method (stdlib, gzip, etc.)

DanielYang59:
> normalizing behaviors across different protocols, e.g., defaulting to "rt" / "utf-8"

For text mode, we're now using UTF-8 by default (as per PEP 597). Regarding `mode` it's a bit hard to argue. I agree `mode='rt'` would be a good enough default (as I guess most user might expect text), but from my side as it wraps around `open/bz2.open/gzip.open/lzma.open`, so I thought it would be better if we require user to explicitly declare what they want (text or binary), as the default could be quite different:
> [open(file, mode='r', buffering=-1, encoding=None, errors=None, newline=None, closefd=True, opener=None)](https://docs.python.org/3/library/functions.html#open)
[bz2.open(filename, mode='rb', compresslevel=9, encoding=None, errors=None, newline=None)](https://docs.python.org/3/library/bz2.html#bz2.open)
[gzip.open(filename, mode='rb', compresslevel=9, encoding=None, errors=None, newline=None)](https://docs.python.org/3/library/gzip.html#gzip.open)
[lzma.open(filename, mode='rb', *, format=None, check=-1, preset=None, filters=None, encoding=None, errors=None, newline=None)](https://docs.python.org/3/library/lzma.html#lzma.open)

Andrew-S-Rosen:
Am I understanding correctly that now the developer must know if the file is `.gz` or plaintext since this will dictate whether they use `mode="rb"` or `mode="rt"`? That seems like it would defeat much of the value of `zopen`, which meant you could just pass in any arbitrary file that may not be known ahead of time and let `zopen` figure out how to handle it with `mode="r"`.

Now it seems one must do:
```python
exts = (".gz", ".GZ", ".bz2", ".BZ2", ".z", ".Z")
mode = "rb" if any(str(logfile_path).endswith(ext) for ext in exts) else "rt"
```
every single time they call `zopen`, which is clunky.

DanielYang59:
I believe it's not the case, now the user/developer should explicitly state the "desired" mode (whether he wants to open the file as text or binary), something like:
```python
from monty.io import zopen

with zopen("test", mode="rt") as f:
    print(f.readlines())
    print(type(f))

with zopen("test.gz", mode="rt") as f:
    print(f.readlines())
    print(type(f))
```

```
['hello\n', 'world\n']
<class '_io.TextIOWrapper'>
['hello\n', 'world\n']
<class '_io.TextIOWrapper'>
```

In other words, one don't need to know beforehand what format that file is in, only need to tell `zopen` **what they want for the final stream**.

---

Previously it was "convenient" with [no guarantee of the stream type](https://github.com/materialyzeai/monty/pull/730), for example:
```python
with zopen("hello.txt") as file:
    print(file.readline())  # "hello" as string

with zopen("hello.txt.gz") as file:
    print((first_line := file.readline()))  # b'hello\n' as bytes
    print("hello" in first_line)  # TypeError: a bytes-like object is required, not 'str'
```

where one might write the same interface and pass different file types which may or may not work, now user just need to explicitly state if they want text or binary to ensure the stream is in their expected type

Andrew-S-Rosen:
Understood, thanks for the clarification! I'll pass that on to the team here.

PR #755, opened 2025-07-24
Include `size` when encoding torch tensor, release `<3.13` pin for `pint` and `torch`

### LINKED ISSUE #679: [Bug]: Dimension of empty Torch tensor is lost during to_json
### Email (Optional)

_No response_

### Version

v2024.4.17

### Which OS(es) are you using?

- [X] MacOS
- [ ] Windows
- [ ] Linux

### What happened?

Essentially, when saving an empty torch tensor of some dimension, that dimensionality information is lost when converting that tensor into json. This should be an easy fix: detect if the tensor is empty, and if so, explicitly save the dimension instead of the data.

```python
import torch

class Klass(MSONable):
    def __init__(self, a):
        self.a = a

k = Klass(torch.empty(0, 2))

print(k.a.shape)
print(k.as_dict())
print(k.to_json())
```

Results in 

```
torch.Size([0, 2])
{'@module': '__main__', '@class': 'Klass', '@version': None, 'a': tensor([], size=(0, 2))}
{"@module": "__main__", "@class": "Klass", "@version": null, "a": {"@module": "torch", "@class": "Tensor", "dtype": "torch.DoubleTensor", "data": []}}
```

Where the dimension of the empty tensor has been lost.

I'd be happy to take a shot at fixing if you'd like.

By the way, why is this important (i.e., why on earth would we care about the dimension of an empty tensor)? Turns out when sampling from the dataless "uninformative" prior using gpytorch, we need to feed it an empty tensor of the correct dimension. I'd like to be able to save this data along with my model!

### Code snippet

```python
See above
```

### Log output

_No response_

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

DanielYang59:
### Summary

- Include shape when encoding torch tensor, to fix #679
- Release `<3.13` pin for `pint` and `torch`
- [x] Add backwards compatibility test


src/monty/json.py
```diff
@@ -597,6 +597,7 @@ def default(self, o) -> dict:
                 "@module": "torch",
                 "@class": "Tensor",
                 "dtype": o.type(),
+                "shape": list(o.shape),
```
DanielYang59, line 600: I bit unsure if we should use `shape` or `size`, looks like they're largely the same thing:
- https://github.com/pytorch/pytorch/issues/5544

However seems `size` might be preferred in some case, e.g.:
- [`torch.empty(*size, *, out=None, dtype=None,...`](https://docs.pytorch.org/docs/stable/generated/torch.empty.html#torch-empty)
- https://docs.pytorch.org/docs/stable/tensors.html:
    - Tensor.shape: Returns the size of the self tensor.
    - Tensor.size: Returns the size of the self tensor.

PR #759, opened 2025-08-17
`ScratchDir` avoid overwriting existing file 

### LINKED ISSUE #758: [Bug]: The `ScratchDir` accidentally overwrites stdout files
### Email (Optional)

_No response_

### Version

2025.3.3

### Which OS(es) are you using?

- [ ] MacOS
- [ ] Windows
- [ ] Linux

### What happened?

As I describe within the context of https://github.com/materialsproject/custodian/issues/389, the `ScratchDir` class has a bug when both `copy_to_current_on_exit=True` and `copy_from_current_on_enter=True` are set.

Consider the scenario where you launch a Slurm job in `/starting/path` for a Python script that calls `ScratchDir`. Upon the Slurm job starting, it will create `/starting/path/slurm-12345.out`. If `copy_from_current_on_enter=True`, it will copy the `slurm-12345.out` file to the temporary scratch directory, run the calculation in the cratch directory, and then copy all the contents of the scratch directory back if `copy_to_current_on_exit=True`. However, the `slurm-12345.out` file is updated in the starting path, not in the scratch directory. This means that when the calculation finishes, the `/starting/path/slurm-12345.out` file is overwritten with the copied file from the start, destroying any logging.

To solve this problem, the `ScratchDir` class should check the timestamps of the files when it copies them back. If the timestamp is older, it should not replace the original file.

### Code snippet

```python

```

### Log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DanielYang59:
fix #758


DanielYang59:
> it is mainly so that you can simplify code where the user may or may not rely on a scratch directory to be made.

I see, this is a valid point

Andrew-S-Rosen:
@DanielYang59 Could you clarify the issue that this PR is solving? I don't quite understand from the title alone.

DanielYang59:
> Could you clarify the issue that this PR is solving? I don't quite understand from the title alone.

Yes sure I'm still working on the implementation and would mark this PR as ready for review later (currently it's still not ready for review)

shyuep:
I actually disagree on this. In the unix command line, copying files from another directory to the current one overwrites the file. There is no check for a newer version. Why would the expectation that this be different here?

